PR #33, opened 2025-11-30
Hotfix 32 bug fixes

karimouf:
T

### DIFF
--- a/backend/webserver/Socket.js
+++ b/backend/webserver/Socket.js
@@ -616,7 +616,6 @@ module.exports = class Socket {
         });
 
         // handle injects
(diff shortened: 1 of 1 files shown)


backend/webserver/Socket.js
```diff
@@ -483,6 +483,21 @@ module.exports = class Socket {
                     d[injection.as] = Number(count.find((c) => c[injection.by] === d.id)?.count) || 0;
                     return d;
                 });
+            } else if (injection.type === "list") {
```
dennis-zyska: As discussed, this could lead to security issues; we should not add this.

PR #50, opened 2025-12-23
[Eiwa] Publish Assessment

akash9676:
Created a new button called "Publish Assessment" and when we click on it we get stepper modal where currently we can download the gradings based on the configuration in the from of CSV.

### New User Features

User can now download gradings in the form of CSV based on the configuration and its related studies that are closed.

### New Dev Features

NA

### Improvements

NA

### Bug Fixes

NA

### Known Limitations

We only have CSV option for now.

### Future Steps
We can integrate the moodle and directly send the gradings to the students and we can also integrate email feature where we can send the gradings via email.

### DIFF
--- a/frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
+++ b/frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
@@ -573,23 +573,16 @@ export default {
       return svc || null;
     },
     /**
-     * Generate assessment key candidate for AI workflow.
-     * Uses format: ${svc.name}_${svc.skill}_assessment
-     */
-    getAssessmentKeyCandidates(studyStep) {
-      const svc = this.getNlpServiceForStudyStep(studyStep);
-      if (!svc || !svc.skill || !svc.name) return [];
-
-      return [`${svc.name}_${svc.skill}_assessment`];
-    },
-    /**
      * Get assessment data key for a study step.
-     * Returns AI key if AI workflow detected, otherwise "assessment_result".
+     * Returns AI key if AI workflow detected (format: ${svc.name}_${svc.skill}_assessment),
+     * otherwise "assessment_result".
      */
     getAssessmentDataKeys(studyStep) {
-      const aiKeys = this.getAssessmentKeyCandidates(studyStep);
-      // If AI workflow detected, only use AI keys (no fallback)
-      return aiKeys.length > 0 ? aiKeys : ["assessment_result"];
+      const svc = this.getNlpServiceForStudyStep(studyStep);
+      if (svc && svc.skill && svc.name) {
+        return [`${svc.name}_${svc.skill}_assessment`];
+      }
+      return ["assessment_result"];
     },
     /**
      * Find the study step for a study that matches the selected configuration.


frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
```diff
@@ -0,0 +1,749 @@
...
+  data() {
+    return {
+      selectedWorkflows: [],
+      selectedConfigurations: [],
+      selectedSessions: [],
+      publishMethod: "csv",
+      linkCollection: "studies",
+    };
+  },
+  computed: {
```
dennis-zyska: Check the computed where we need to filter for the selected projects (search code for reference). Especially if we have several projects, we want to show the data only for the current selected project.

frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
```diff
@@ -0,0 +1,749 @@
...
+      return this.selectedConfigurations.length > 0 ? this.selectedConfigurations[0].id : null;
+    },
+    selectedConfigurationContent() {
+      if (!this.selectedConfigurationId) return null;
+      const config = this.$store.getters["table/configuration/get"](this.selectedConfigurationId);
+      return config?.content || null;
+    },
+    selectedConfigurationName() {
+      if (!this.selectedConfigurationId) return "unknown";
+      const config = this.$store.getters["table/configuration/get"](this.selectedConfigurationId);
```
dennis-zyska: you request the vuex stores with the same command twice (also in `selectedConfigurationContent`). Create a new computed `selectedConfiguration()` where you request the vuex store and use it in other functions.

frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
```diff
@@ -0,0 +1,749 @@
...
+            firstName: user?.firstName || "Unknown",
+            lastName: user?.lastName || "Unknown",
+            userName: user?.userName || "-",
+            ownerFirstName: owner?.firstName || "-",
+            ownerLastName: owner?.lastName || "-",
+            ownerUserName: owner?.userName || "-",
+            ownerExtId: owner?.extId || "",
+            submissionId: submission?.id || null,
+            submissionExtId: submission?.extId || "",
+            link: window.location.origin + "/review/" + session.hash,
```
dennis-zyska: This was already clarified right? We should not use the link, instead better just use the hash. The links are only needed in the publishing of the links in Moodle.

frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
```diff
@@ -0,0 +1,749 @@
...
+      // Find any service with a skill (name and type can be anything)
+      const svc = cfg.services.find((s) => s.skill) || cfg.services[0];
+
+      return svc || null;
+    },
+    /**
+     * Generate assessment key candidate for AI workflow.
+     * Uses format: ${svc.name}_${svc.skill}_assessment
+     */
+    getAssessmentKeyCandidates(studyStep) {
```
dennis-zyska: All those functions, looks like a long way to get the column keys. We should do this more easy, just use the union of all dictionary keys of the selected `document_data` assessment results. that way we also make sure that we don't miss data.

PR #60, opened 2026-01-07
Feat 58 submission workflow

karimouf:
## **Main Description**
A new revision‑based submission review workflow has been introduced to the CARE system. This feature allows administrators and reviewers to create new submissions that build on data from previous study sessions. When reviewing a revised submission, the previous version is displayed in read‑only mode, while the new version appears with the assessment sidebar pre‑filled using the prior submission's grade (without persisting it). A new assignment mode ensures that submissions can only be created when all selected users have an associated study session.

---

## **New User Features**
- Ability to create a new study based on existing study session data with workflow step mapping to previous submissions
- Read‑only display of the previous submission version during revision review via `readOnlyComponents` configuration
- Assessment sidebar in the new submission pre‑populated with the previous submission's grade from `previous_assessment_result` (not saved automatically, marked with `isSaved: false`)
- New assignment mode `study_session` requiring all selected users to have a corresponding study session before submission
- Clear error messaging when users lack a required study session or when the previous submission cannot be found

---

## **New Dev Features**
- Added `previousSubmissionId` column to `submission` table (migration 20260103134710-extend-submission-previousSubmissionId.js)
- Added `parentSubmissionId` foreign key to track submission lineage and revision chains
- New `Revision Workflow` with two-step annotation process added to workflow system (migration 20260103145255-extend-workflow-revision_workflow.js)

---

## **Improvements**
- Handled multiple pdfs being loaded in the same study
---

## **Bug Fixes**
- Fixed document data duplication to create new entries instead of modifying original document data
- Fixed assessment component based readonly
---

## **Known Limitations**
- Cannot create a new document_data entry for the same study session

---

## **Future Steps**
NA

### DIFF
--- a/backend/webserver/sockets/document.js
+++ b/backend/webserver/sockets/document.js
@@ -815,8 +815,9 @@ class DocumentSocket extends Socket {
                 if (!validationResult.success) {
                     throw new Error(validationResult.message || "Validation failed");
                 }
+                // 3. Get previous submission for the user and project to link the new submission (if exists)
                 const previousSubmission = await this.models["submission"].getParentSubmission(submission.userId, submission.projectId, true, {transaction});
-                // 3. Only if validation passes, create submission and save documents
+                // 4. Only if validation passes, create submission and save documents
                 const submissionEntry = await this.models["submission"].add(
                     {
                         userId: submission.userId,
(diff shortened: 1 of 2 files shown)


backend/db/models/submission.js
```diff
@@ -68,28 +68,49 @@ module.exports = (sequelize, DataTypes) => {
...
+        /**
+         * Get the most recent submission for a user and project, traversing back through parent submissions
+         * 
+         * @param {number} userId Id of the user
+         * @param {number} projectId Id of the project
+         * @param {boolean} root Whether to return the root submission
+         * @param {Object} transaction Database transaction object
+         * @returns {Object|null} The most recent submission or null if none found
+         */
+        static async getPreviousRootSubmission(userId, projectId, root, transaction = {}) {
```
dennis-zyska: lets define it as getParentSubmission(userId, projectId, returnRoot, transaction = {})

PR #65, opened 2026-01-20
Apply Skills

akash9676:
Apply Skills Processing - Completion Confirmation & Error Tracking

### New User Features
Processing completion now requires admin confirmation before clearing state. Errors encountered during processing are displayed with timestamps and context. Timeout warning shown when a request takes longer than expected

### New Dev Features
`confirmCompletion` command added to BackgroundTaskService. `disconnectClient` method for automatic subscription cleanup

### Improvements
"View Processing" button remains visible after completion until confirmed. Real-time subscription/unsubscription added to Submissions dashboard. Processing stepper shows "Confirm" button when completed instead of "Cancel"

### Bug Fixes
Fixed subscription cleanup when clients disconnect. Cancelled processes now clear state immediately instead of showing as completed

### Known Limitations
NA

### Future Steps
- N/A

### DIFF
--- a/backend/webserver/services/backgroundTask.js
+++ b/backend/webserver/services/backgroundTask.js
@@ -1,4 +1,5 @@
 const Service = require("../Service.js");
+const Socket = require("../Socket.js");
 const fs = require("fs");
 const path = require("path");
 const UPLOAD_PATH = `${__dirname}/../../../files`;
@@ -176,7 +177,7 @@ module.exports = class BackgroundTaskService extends Service {
             currentSubmissionsCount: 0,
             currentRequestId: null,
             batchStartTime: Date.now(),
-            errors: []  // Track all errors that occurred
+            errors: [],  // Track all errors that occurred
             nlpTimeout
         };
         this.sendAll("backgroundTaskUpdate", this.backgroundTask);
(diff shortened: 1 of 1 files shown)


backend/webserver/services/backgroundTask.js
```diff
@@ -144,16 +163,21 @@ module.exports = class BackgroundTaskService extends Service {
...
+
         this.backgroundTask.preprocess = {
             cancelled: false,
+            completed: false,
             requests: {},
             currentSubmissionsCount: 0,
             currentRequestId: null,
-            batchStartTime: Date.now()
+            batchStartTime: Date.now(),
+            errors: []  // Track all errors that occurred
```
karimouf: comma expected here ...

frontend/src/basic/modal/ApplySkillModal.vue
```diff
@@ -16,6 +16,7 @@
         :show-close="true"
         cancel-next-text="Cancel Preprocess"
         @cancel="cancelProcessing"
```
karimouf: I see process active or completed are handled, what about if it's cancelled we only show a pop up should it maybe require extra handling?

PR #71, opened 2026-01-27
Bug fixes issue #10 

### LINKED ISSUE #10: [BUG] Error when document is not loaded
# Describe the Bug:  

If the annotator cannot load the document because it is unavailble (file does not exist), it should show an error message to the user.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

N/A

# Expected Behavior:  

When opening a document in the annotator and the file was removed beforehand, currently it only shows a white page and there is no information for the user. The user should receive an error after some time if the document is not loaded.

# Actual Behavior:  

N/A

# Screenshots:  

N/A

# System Information: 

N/A

# Additional Information:  

N/A

ututono:
**Goal:**  This patch aims to resolve the issue https://github.com/UKPLab/CARE/issues/10, addressing multiple bugs related to document validation. The main focus is on providing a better user experience when documents are missing, deleted, or inaccessible. 

**Key Changes:**
1. Added a new `validateDocument()` method in `backend/webserver/sockets/document.js` that centralizes document validation logic:
    - Checks for: database record existence, user access permissions, and file existence on disk
    - Returns structured error codes (`DOCUMENT_NOT_FOUND`, `ACCESS_DENIED`, `FILE_MISSING`), details are in the comment https://github.com/UKPLab/CARE/issues/10#issuecomment-3805296245,  for consistent error handling

2. Improved UX - Error Pages Instead of Toasts. Replaced toast notifications with more informative error pages when documents are unavailable. Implemented in both `Document.vue` and `Study.vue ` components.
3. Loading Modal Fixes. Fixed issue where the loading modal would freeze when encountering document errors.  The result why `close()` can not correctly terminate the loading modal seems to be` hide()` deferred when `isShown=false`, but component unmounts before deferred hide runs in` frontend/src/basic/Modal.vue`.  Modal now properly closes and emits error events when a document cannot be loaded by replacing `hide()` with `dispose()` in `beforeunmount()`.
4. Error messages are now consistent whether a file was deleted through the UI or is missing from the file system.

**Demonstration:**
Test accessing shared links to the document which has been deleted from the UI interface and from the filesystem accordingly. 

Test accessing shared links to the study which uses a document which has been deleted from the UI interface and from the filesystem accordingly.

### DIFF
--- a/backend/webserver/sockets/app.js
+++ b/backend/webserver/sockets/app.js
@@ -322,7 +322,7 @@ class AppSocket extends Socket {
 
         if (result.length === 0) {
             // Record exists but user doesn't have permission
-            const error = new Error("You don't have permission to access this resource.");
+            const error = new Error("You don't have rights to access this data.");
             error.errorCode = "ACCESS_DENIED";
 
             // Emit specific error event for study/session tables
--- a/backend/webserver/sockets/app.js
+++ b/backend/webserver/sockets/app.js
@@ -289,12 +289,58 @@ class AppSocket extends Socket {
      * @throws {Error} Throws error if results is empty (no record found or operation fails)
      */
     async sendDataByHash(data, options) {
+        const record = await this.models[data.table].getByHash(data.hash);
+
+        if (!record) {
+            // Record doesn't exist or was deleted
+            const error = new Error("The requested resource does not exist or has been deleted.");
+            error.errorCode = "NOT_FOUND";
+
+            // Emit specific error event for study/session tables
+            if (data.table === "study") {
+                this.socket.emit("studyError", {
+                    errorCode: "NOT_FOUND",
+                    message: error.message,
+                    studyHash: data.hash
+                });
+            } else if (data.table === "study_session") {
+                this.socket.emit("studySessionError", {
+                    errorCode: "NOT_FOUND",
+                    message: error.message,
+                    studySessionHash: data.hash
+                });
+            }
+
+            throw error;
+        }
+
+        // Now check permissions by attempting to send via filtered sendTable
         const result = await this.sendTable(data.table, mergeFilter([[{
             key: "hash",
             value: data.hash
         }]], this.models[data.table].getAttributes()));
+
         if (result.length === 0) {
-            throw new Error("You don't have rights to access this data");
+            // Record exists but user doesn't have permission
+            const error = new Error("You don't have permission to access this resource.");
+            error.errorCode = "ACCESS_DENIED";
+
+            // Emit specific error event for study/session tables
+            if (data.table === "study") {
+                this.socket.emit("studyError", {
+                    errorCode: "ACCESS_DENIED",
+                    message: error.message,
+                    studyHash: data.hash
+                });
+            } else if (data.table === "study_session") {
+                this.socket.emit("studySessionError", {
+                    errorCode: "ACCESS_DENIED",
+                    message: error.message,
+                    studySessionHash: data.hash
+                });
+            }
+
+            throw error;
         }
     }
 
--- a/backend/webserver/sockets/app.js
+++ b/backend/webserver/sockets/app.js
@@ -290,28 +290,12 @@ class AppSocket extends Socket {
      */
     async sendDataByHash(data, options) {
         const record = await this.models[data.table].getByHash(data.hash);
+        let errorCode = "UNKNOWN";
 
         if (!record) {
             // Record doesn't exist or was deleted
-            const error = new Error("The requested resource does not exist or has been deleted.");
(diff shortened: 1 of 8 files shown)


backend/webserver/sockets/app.js
```diff
@@ -289,12 +289,58 @@ class AppSocket extends Socket {
...
+        const record = await this.models[data.table].getByHash(data.hash);
+
+        if (!record) {
+            // Record doesn't exist or was deleted
+            const error = new Error("The requested resource does not exist or has been deleted.");
+            error.errorCode = "NOT_FOUND";
+
+            // Emit specific error event for study/session tables
+            if (data.table === "study") {
+                this.socket.emit("studyError", {
```
dennis-zyska: We don't need to send extra websockets to the frontend, we can use the callback, see [`createSocket`](https://github.com/UKPLab/CARE/blob/main/backend/webserver/Socket.js#L59C5-L59C17). We can just throw an error and have the information already in frontend, like  https://github.com/UKPLab/CARE/blob/main/backend/webserver/sockets/comment.js#L29 and https://github.com/UKPLab/CARE/blob/main/frontend/src/components/annotator/sidebar/card/Comment.vue#L473. So also no need for listening to the sockets in frontend.

dennis-zyska (changes requested):

dennis-zyska (changes requested):
The solution with the text parsing with the error codes should be improved. We can change the function here: https://github.com/UKPLab/CARE/blob/main/backend/webserver/Socket.js#L86 and add a specific variables for the error code. It should be backward compatible. One possible solution would be to create an Error from the Error class and set the error code. If the error code is set, we send it back to the frontend, otherwise we use the previous solultion (in that way we don't need to adapt to much from the existing code). Here is a potential solution that could work: https://stackoverflow.com/a/69166148

frontend/src/components/study/LoadingModal.vue
```diff
@@ -353,6 +364,18 @@ export default {
 
       this.close();
     },
+
+    getErrorMessage(response) {
```
dennis-zyska: We define here the message and codes again, does this make sense? If we send all the data already from the backend to the frontend?

dennis-zyska (changes requested):

backend/webserver/sockets/app.js
```diff
@@ -289,12 +289,32 @@ class AppSocket extends Socket {
      * @throws {Error} Throws error if results is empty (no record found or operation fails)
      */
     async sendDataByHash(data, options) {
+        const record = await this.models[data.table].getByHash(data.hash);
+        let errorCode = "UNKNOWN";
+        let errorMessage = "";
+
+        if (!record) {
+            // Record doesn't exist or was deleted
+            errorCode = "NOT_FOUND";
```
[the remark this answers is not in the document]
ututono: I introduced a reusable `generateError(code, message)` helper in` utils/generic.js` and refactored this branch to use `throw generateError(...)` here (and in similar socket paths)

frontend/src/components/study/LoadingModal.vue
ututono: I removed the duplicated local error-message map in `LoadingModal` and same function in other place such as `Document.vue` and now use backend-provided `response.message` directly, but I introduced a map `titleMap` between error code and title of the loading modal to provide more readable information in frontend.

ututono:
@dennis-zyska I’ve addressed your requested changes and pushed an updated [commit](https://github.com/UKPLab/CARE/pull/71/commits/43a3b33ea1b9a9a0f1318630efa2016d1adf6bc2). And also merged the latest dev into this bug fix branch. Could you please take another look?

@NilsDy FYI

Also this PR could be reviewed within my team first, I'd assign it to one of team members for first review

backend/webserver/sockets/document.js
```diff
@@ -1097,6 +1130,8 @@ class DocumentSocket extends Socket {
      * @returns {Promise<Object>} A promise that resolves with the retrieved `document_data` record object from the database.
      */
     async getDocumentData(data, options) {
+        const docuemt = await this.validateDocument(data.documentId, 'id', true);
```
dennis-zyska, line 1133: variable name: document

dennis-zyska (approved):

theultimatewasteofspace:
Just went through the changes. Centralizing everything into validateDocument() seems to be a much cleaner approach than handling it separately everywhere. The generateError() helper is also a nice addition. And the hide() to dispose() fix for the loading modal makes sense once you see why close() wasn't working properly. Overall looks good to me!

PR #89, opened 2026-02-15
Feat 21 Two Factor Authentication

### LINKED ISSUE #21: [FEATURE] Two-factor Authentication

To increase security, we should add two factor authentication.

Add basic two factor authentication (send mail with token to login).

none

## Additional context 

Refer to the following resources:

- https://www.passportjs.org/packages/passport-orcid/
- https://www.passportjs.org/packages/passport-ldapauth/
- https://www.passportjs.org/packages/passport-local-token/ (for mail, maybe)
- For totp, there are two packages: https://www.passportjs.org/packages/passport-2fa-totp/ & https://www.passportjs.org/packages/passport-totp/

pdji1602003:
## Main Description

This merge request introduces 2fa implementation and third-party login methods.

## New User Features

1. User can go to user menu > Configure Authentication to configure 2fa and enhance account security.
2. CARE offers one time token via email and TOTP for 2fa.
3. If user activates both 2fa methods (email & TOTP), user will be directed to a select page after login.
4. CARE offers three extra third-party login flows: orcid, LDAP, and SAML.
5. Admin can enable these login flows in the Setting dashboard.
6. Admin can enforce 2fa in the Setting dashboard.
7. Terms description about email processing was updated as we now use email to send the one time code.

### Note
1. SAML login flow requires testing.

### Known Limitations
1. Sever needs to be manually restarted if the admin changes any setting about third-login  methods (orcid, LDAP, and SAML)
2. In the 2fa enforcement flow, everyone is impacted, including admins. Consider admin exemption in the 2fa enforcement flow.
3. Old user data is cleared now via page reload. A data clearing mechanism is needed.
4. There is no page where user can provide their email.


backend/db/models/user.js
```diff
@@ -578,6 +555,20 @@ module.exports = (sequelize, DataTypes) => {
...
+            twoFactorOtpExpiresAt: DataTypes.DATE,
+            // Multi-method 2FA configuration
+            twoFactorMethods: {
+                type: DataTypes.JSON,
+                defaultValue: [],
+            },
+            // TOTP for 2FA
+            totpSecret: DataTypes.STRING,
+            // External login method identifiers
+            orcidId: DataTypes.STRING,
```
melolw: [question] should we add unique: true here too?

melolw (commented):
I also have a general question because I haven't seen something like limiting the amount to tries? Like if someone tries to brute force all 6 digits possibilities, is there something stopping them?

Overall the code is great! I learned a lot along the way, I have limited knowledge about authentification so most of my questions are things I saw in theory only. I also think it's a great addition to the software, 2FA is such an important feature 😎

backend/db/models/user.js
pdji1602003: We usually set the extra rules in the migration files; here we do not need to add `unique: true`.

frontend/src/auth/TwoFactorSelect.vue
```diff
@@ -0,0 +1,249 @@
...
+                  </div>
+                </div>
+              </div>
+            </div>
+
+            <div class="col-md-6 offset-md-4 my-4">
+              <button
+                class="btn btn-primary btn-block"
+                type="submit"
+                :disabled="!selectedMethod || isSubmitting"
```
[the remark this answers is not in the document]
pdji1602003: This is a very cautious suggestion, but I am not sure if it is necessary. The user will be directed to the next page as soon as they click on "Continue" button. There is little to no time for them to click “Cancel.”

frontend/src/auth/TwoFactorSelect.vue
```diff
@@ -0,0 +1,249 @@
...
+  created() {
+    this.initMethodsFromQuery();
+  },
+  methods: {
+    initMethodsFromQuery() {
+      const raw = this.$route.query.methods;
+      if (typeof raw === "string" && raw.length > 0) {
+        this.methods = raw.split(",").filter((m) => !!m);
+      }
+      // Fallback: just support email/totp if nothing present
```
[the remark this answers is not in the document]
pdji1602003: Users are directed here only when they have two 2FA methods, so the fallback should be email/totp as well; otherwise, they wouldn't be on this page, but TwoFactorVerifyEmail or TwoFactorVerifyTotp.

backend/utils/auth.js
```diff
@@ -127,3 +127,13 @@ exports.decodeToken = function decodeToken(token) {
...
     }
 }
+
+/**
+ * Generate a 6-digit OTP (One-Time Password)
+ * @returns {string} 6-digit OTP
+ */
+exports.generateOTP = function generateOTP() {
+    // Generate a random 6-digit number (000000 to 999999)
+    const otp = Math.floor(100000 + Math.random() * 900000).toString();
```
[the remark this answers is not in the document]
pdji1602003: This is a good catch! Yes, since we use Node.js, it's better to use its module. I have updated it in commit 3c256452

backend/db/migrations/20260126144237-extend-user.js
```diff
@@ -0,0 +1,79 @@
...
+    type: "STRING",
+    allowNull: true,
+    defaultValue: null,
+  },
+];
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.changeColumn("user", "email", {
```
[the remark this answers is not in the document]
pdji1602003: Resolved in commit ee743fc58637f7f8770eb907865f104ed6a5ee88

We are allowing email to be Null now because we added OrcId login method this time, which means users can log in using their orcid account. Orcid users can choose not to share their email with our platform. In this case, we cannot get such information. If we are not allowing email to be null, then we will run into issue when creating user account for orcid users.

frontend/src/auth/TwoFactorVerifyEmail.vue
```diff
@@ -0,0 +1,297 @@
...
+          },
+        );
+
+        if (response.status === 200) {
+          this.showSuccess = true;
+          this.successMessage =
+            "A new verification code has been sent to your Email";
+
+          // Start cooldown timer (60 seconds)
+          this.resendCooldown = 60;
```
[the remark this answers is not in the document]
pdji1602003: Resolved in commit in commit 12d73e97.

backend/db/migrations/20260223125239-extend-setting-auth_methods_2fa.js
```diff
@@ -0,0 +1,136 @@
...
+    {
+        key: "system.auth.orcid.clientId",
+        type: "string",
+        value: "",
+        description: "ORCID OAuth client id (changes require a restart of the server)",
+        onlyAdmin: true
+    },
+    {
+        key: "system.auth.orcid.clientSecret",
+        type: "string",
```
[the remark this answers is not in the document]
pdji1602003: Database encryption will be left with Junaid to handle.

pdji1602003:
> I also have a general question because I haven't seen something like limiting the amount to tries? Like if someone tries to brute force all 6 digits possibilities, is there something stopping them?

You are right! I have implemented the rate limit to prevent this from happening.

backend/webserver/routes/auth.js
```diff
@@ -76,11 +120,387 @@ module.exports = function (server) {
...
+    }
+
+    async function isLoginMethodEnabled(method) {
+        return (await server.db.models['setting'].get(`system.auth.${method}.enabled`)) === "true";
+    }
+
+    const EMAIL_2FA_RESEND_COOLDOWN_MS = 60 * 1000;
+    const MAX_2FA_VERIFY_ATTEMPTS = 5;
+
+    function getEmailOtpCooldownInfo(pending) {
```
dennis-zyska: Here are a few functions without a docstring.

backend/webserver/routes/auth.js
```diff
@@ -76,11 +120,387 @@ module.exports = function (server) {
...
+
+            delete req.session.twoFactorPending;
+
+            return req.session.save((err) => {
+                if (err) {
+                    server.logger.error("Failed to save session: " + err);
+                    return res.status(500).json({ message: "Session error during 2FA." });
+                }
+
+                return res.status(429).json({
```
dennis-zyska: Can we also add here `RetryAfter` information? (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) Or is it final? If it is final, the 429 error would not apply, and I think it should be possible again after some time, right?

backend/webserver/routes/auth.js
```diff
@@ -76,11 +120,387 @@ module.exports = function (server) {
...
+
+        await server.db.models['user'].update(
+            {
+                twoFactorOtp: otp,
+                twoFactorOtpExpiresAt: otpExpiresAt
+            },
+            { where: { id: userRecord.id } }
+        );
+
+        await server.sendMail(
```
dennis-zyska: We should use the email template feature we already integrated. Please have a look here: https://github.com/UKPLab/CARE/blob/7cdc0ed137db2a5b0da021bd4de485e66682a68e/backend/webserver/routes/auth.js#L295
(We don't want to save the mail in the source code, it should be in the database and as file on disk for fallback)

But you can also ask @mohammadsherif0

backend/webserver/routes/auth.js
```diff
@@ -76,11 +120,387 @@ module.exports = function (server) {
...
+     * updating login records, and handling the HTTP response.
+     * @param {Object} req - The Express request object.
+     * @param {Object} res - The Express response object.
+     * @param {Object} user - The user object to be logged in.
+     * @param {Object} options - Configuration for the response.
+     * @param {string} [options.mode='json'] - The response mode: 'json' or 'redirect'.
+     * @param {string} [options.redirectPath='/dashboard'] - The path to redirect to if mode is 'redirect'.
+     * @returns {Promise<void>}
+     */
+    async function finalizeLogin(req, res, user, options = { mode: 'json', redirectPath: '/dashboard' }) {
```
dennis-zyska: What if we open a study URL before logging in to the account? Are we still redirected to that study URL after successful login

backend/webserver/routes/auth.js
dennis-zyska, line 1: This is a huge file, I think we can create a subfolder `./backend/webserver/routes/auth/...` and create subfiles for examples `./auth/totp.js` for each method.

backend/webserver/Server.js
```diff
@@ -21,6 +24,69 @@ const RPC = require(path.resolve(__dirname,"./RPC.js"));
...
+        };
+    }
+
+    const parts = (fullName || "").trim().split(/\s+/).filter(Boolean);
+    return {
+        firstName: normalizedFirstName || toDisplayNamePart(parts[0], fallbackFirstName),
+        lastName: normalizedLastName || toDisplayNamePart(parts.slice(1).join(" "), fallbackLastName),
+    };
+}
+
```
dennis-zyska: The docstrings are missing, can we move this block to some files like `../utils/auth.js` , this would be much cleaner

backend/webserver/Server.js
```diff
@@ -205,42 +279,323 @@ module.exports = class Server {
...
 
-            crypto.pbkdf2(password, user.salt, 310000, 32, 'sha256', (err, hashedPassword) => {
-                if (err) {
-                    return cb(err);
-                }
+        // 2. Initialize Strategies
+        await this.#setupLocalStrategy();
+        await this.#setupOrcidStrategy();
+        await this.#setupLdapStrategy();
+        await this.#setupSamlStrategy();
```
dennis-zyska: Maybe we can move those functions as well into an extra file for the loginManagement (like a class Login.js that holds everything that is important regarding the Login). This file is also already really big and it would make it easier to get

frontend/src/router.js
```diff
@@ -33,6 +33,30 @@ const routes = [
...
+        path: "/login/ldap",
+        component: () => import("@/auth/LoginLdap.vue"),
+        name: "login-ldap",
+        meta: {requireAuth: false, hideTopbar: true, checkLogin: true}
+    },
+    {
+        path: "/2fa/verify/email",
+        name: "2fa-verify-email",
+        component: () => import("@/auth/TwoFactorVerifyEmail.vue"),
+        meta: { requiresAuth: false }
```
dennis-zyska: What is with the additional meta information like hideTopbar and checkLogin, also in the other new routes?

dennis-zyska (changes requested):
Really nice implementation, the code is already in a really good shape, but there are some smaller issues, like a missing docstring. It could also be rearranged a bit to make the code more understandable.

frontend/src/router.js
pdji1602003: Resolved in commit 62ed88ae.

backend/webserver/Server.js
pdji1602003: Resolved in commit 7cda2748.

backend/webserver/Server.js
pdji1602003: Resolved in commit https://github.com/UKPLab/CARE/commit/7cda27485c469ecd20b94e40802ac4be17237d20.

backend/webserver/routes/auth.js
pdji1602003: Resolved in commit https://github.com/UKPLab/CARE/commit/1e959d5d63fc70e4811ef8413c53dcf4ed747a9d.

backend/webserver/routes/auth.js
pdji1602003: Your observation was right. There wasn't the redirect mechanism in place previously.
Resolved it in commit 8bb77770.

backend/webserver/routes/auth.js
pdji1602003: Resolved it in commit a4c15335.

backend/webserver/routes/auth.js
pdji1602003: Add RetryAfter info in commit 0b9feb50.

backend/webserver/routes/auth.js
pdji1602003: Resolved it in commit 1e959d5d63fc70e4811ef8413c53dcf4ed747a9d.

backend/webserver/routes/auth.js
pdji1602003: @mohammadsherif0

I check the template related feature and there are a number of things that feel confusing (but maybe I just don't understand this feature well enough):

1. the return value of getEmailContent contains fallback.subject, which means this feature couldn't work without fallback. In this sense, it's not a fallback mechanism, it's a dependence, right? I wonder why we didn't save the subject information into the DB, but depend on this two-step procedure of retrieving information required for an email.
https://github.com/UKPLab/CARE/blob/6945b3b3e6f95faf44ea0cf36bb1127d26782cef/backend/utils/emailHelper.js#L78
2. When I tried to create a new system email, I needed to check several files, modify several places such as templateResolver, create migration, and create fallback files. The procedure feels a bit fragmented. Is it intended like this? 
3. It’s related to the previous point. if the new system email requires new kinds of placeholder, we need to go through migration, check frontend (to if the frontend component supports this placeholder) and may need to expand the backed logic. It feels a bit complicated from my perspective.
4. The placeholder table has the column, type, but it does not mean the type of the placeholder, but the type of template, which makes me feel like it’s a foreignKey, while in reality, it isn’t. The naming is a bit confusing for two reasons:
    1. If it’s not a foreign key, then there is no restraint upon it. It could be a type that does not exist in Template table and the database wouldn’t give us warning.
    2. I would actually assume that the type of placeholder takes value such as ‘text’ (what placeholderType does/holds now)and we have another column called, say, templateType. 
5. On the frontend, the Publish Template feature shows a list of templates published by others, excluding our own public templates. I know there is description explaining this, but I kind of feel it’s not intuitive if we exclude public templates from our own.

backend/webserver/Server.js
pdji1602003: To add more context, I organized the strategies as functions rather than a class because this layer is mostly stateless startup logic and only gets initialized once when the server starts up. But if you think this is not a good idea, please let me know.

dennis-zyska (approved):

PR #110, opened 2026-03-04
Feat 44 template dashboard

mohammadsherif0:
### **Description**

This feature delivers a template-management flow, where users can create/edit/publish/copy/update templates with clear ownership and visibility rules, template placeholders are resolved reliably, and template content is consistently handled across frontend and backend paths. It includes the full lifecycle from creation and content editing to publishing and source-based copy updates, while keeping template-specific constraints enforced.

### **New User Features** 

- Users can create and manage templates through the dashboard.
- Templates support multilingual content and default language handling.
- Templates can be published for reuse and copied by other users.
- Template copies can be updated from their source template.
- Placeholder-based content is resolved for runtime use.

### **New Dev Features**

- Added dedicated backend template models to separate template metadata, stable multilingual content, and draft edits for cleaner maintenance.
- Implemented template socket API for lifecycle actions.
- Moved template row updates to shared appDataUpdate where applicable, while enforcing ownership/copy/public rules in model hooks.
- Centralized template placeholder resolution and fallback handling so runtime output behavior is consistent across template users.
- Extracted reusable form rendering into FormFields component so regular and advanced BasicForm sections use one shared implementation.
- Added Collapsible component usage for advanced settings grouping so section behavior is consistent and reusable.

### **Known Limitations**

- Template assignments in Settings currently require a backend restart to be fully picked up in runtime email/template usage.
- Template editing still relies on a template-specific editor wrapper flow, which introduces overlap with the generic `Editor.vue` logic and increases maintenance complexity.

### **Future Steps**

- Refactor template editing architecture to reduce duplication between template-specific editor code and `Editor.vue`: either move template-mode branching one level up into `Template.vue`, or integrate required template behavior directly into the shared `Editor.vue` while keeping template-only concerns minimal.


backend/db/migrations/20251227183831-create-template-placeholder-mapping.js
dennis-zyska, line 1: I think we can generalize this db table a bit, for now it is just use in templates, but we will have also placeholder in "Modal" documents, so why not naming the db `placeholder` instead of `template_placeholder_mapping` and the column `type` instead of `templateType`.

Please also think to adapt the name of the migration.

backend/db/models/placeholder.js
dennis-zyska, line 1: as described in my previous comment, should be generalized to `placeholder`

backend/utils/emailHelper.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
+const { resolveTemplate } = require("./templateResolver");
+
+/**
+ * Get email content from template or fallback to hardcoded text
+ * (Same pattern as auth.js getEmailContent)
```
dennis-zyska, line 35: add author information

backend/db/migrations/20260130202032-create-template-content.js
dennis-zyska, line 1: we can name the db `template_content` instead of `template_language_content`

frontend/src/components/Template.vue
```diff
@@ -0,0 +1,111 @@
...
+      },
+    },
+    data() {
+      return {
+        templateIdNum: 0
+      }
+    },
+    computed: {
+      template() {
+        return this.$store.getters["table/template/get"](this.templateIdNum);
```
dennis-zyska, line 42: When I directly open this route, where does the Vuex Store is getting the db entries from? I think we need a subscription to the template table here

backend/db/models/template_content.js
dennis-zyska, line 1: see previous comment on change the db name

backend/db/models/template.js
```diff
@@ -0,0 +1,418 @@
...
+                sourceId: sourceTemplateId,
+            };
+
+            const copiedTemplate = await Template.add(
+                Object.assign(baseData, overrides, { force: undefined }),
+                { transaction }
+            );
+
+            // Copy all template_language_content rows
+            await Template.copyLanguageContent(source.id, copiedTemplate.id, options);
```
dennis-zyska, line 235: Does it make sense to copy the edits here as well? How do you merge new edits with the existing content?

backend/db/models/template.js
```diff
@@ -0,0 +1,418 @@
...
+         * Updates existing rows in place (or adds if missing) to respect UNIQUE(templateId, language).
+         *
+         * @param {number} copyId - The ID of the copied template
+         * @param {Object} options - Database options including transaction
+         * @returns {Promise<Object>} The updated copy
+         */
+        static async updateFromSource(copyId, options = {}) {
+            const { Op } = require("sequelize");
+            const transaction = options.transaction;
+            const Tlc = sequelize.models.template_language_content;
```
dennis-zyska: Please don't use short variable names like this. It saves space and characters, but it makes the code hard to read and follow. In general, variables should be longer and self-explanatory.

backend/db/models/study_step.js
```diff
@@ -105,6 +106,15 @@ module.exports = (sequelize, DataTypes) => {
                     ? sequelize.models.document.docTypes.DOC_TYPE_HTML
                     : sequelize.models.document.docTypes.DOC_TYPE_MODAL;
 
+                // Check if documentId is a template selection (format: "template:123")
+                if (typeof data.documentId === 'string' && data.documentId.startsWith('template:')) {
```
dennis-zyska: I'm not that happy with a string in documentId, can't we save an additional value in data directly, and check if it exists?

backend/db/models/study_step.js
```diff
@@ -170,11 +182,31 @@ module.exports = (sequelize, DataTypes) => {
...
+                            
+                            const resolvedDelta = await resolveTemplateToDelta(
+                                options.context.documentTemplateId,
+                                {
+                                    baseUrl: baseUrl
+                                },
+                                sequelize.models,
+                                options
+                            );
+                            const deltaFilePath = path.join(UPLOAD_PATH, `${newDocument.hash}.delta`);
```
dennis-zyska: here we need to make sure that we also save the edits, some mechanism don't depend only on the delta file, also that the edits exists. So we should never "forget" about the edits, even if we already combined the data in the delta. Maybe we can directly focus on the edits, set the document_edits entries for draft on True and let the existing code handle the delta disk files. The document_edit are always the ground truth, not the delta file. Delta file is simply for caching.

backend/utils/templateResolver.js
```diff
@@ -0,0 +1,344 @@
...
+ */
+const Delta = require("quill-delta");
+
+/**
+ * Extract plain text from Quill Delta operations
+ * 
+ * @param {Object} delta - Quill Delta object with ops array
+ * @returns {string} Plain text extracted from Delta
+ */
+function extractTextFromDelta(delta) {
```
dennis-zyska, line 18: We already have code for translating db --> delta and delta --> db here: https://github.com/UKPLab/CARE/blob/main/utils/modules/editor-delta-conversion/index.js

We should reuse this code as much as possible instead of creating an additional one, to make sure we don't introduce new potential errors

backend/utils/templateResolver.js
```diff
@@ -0,0 +1,344 @@
...
+ * Falls back to template.defaultLanguage if the requested language has no row.
+ *
+ * @param {number} templateId - Template ID
+ * @param {string} language - Language code (e.g. 'en', 'de')
+ * @param {Object} models - Database models object
+ * @param {Object} options - Options (e.g. transaction)
+ * @returns {Promise<Object>} Content object with ops array, or null if no row exists
+ */
+async function getTemplateContentForLanguage(templateId, language, models, options = {}) {
+    const Tlc = models["template_language_content"];
```
dennis-zyska: Here as well, don't use abbreviations for variable names; directly use the variables to increase readability

backend/utils/templateResolver.js
```diff
@@ -0,0 +1,344 @@
...
+    let text = extractTextFromDelta(originalDelta);
+    let resolvedText = text;
+    
+    for (const [placeholder, value] of Object.entries(replacements)) {
+        const escapedPlaceholder = placeholder.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        const regex = new RegExp(escapedPlaceholder, 'g');
+        resolvedText = resolvedText.replace(regex, value || "");
+    }
+    
+    const resolvedDelta = new Delta();
```
dennis-zyska, line 284: reuse existing code: https://github.com/UKPLab/CARE/blob/main/utils/modules/editor-delta-conversion/index.js

If you need new code, think about adding it to this additional module if it is just about the delta / db conversion

backend/webserver/sockets/assignment.js
```diff
@@ -547,6 +587,65 @@ class AssignmentSocket extends Socket {
...
+
+        // Get baseUrl from settings
+        const baseUrl = await this.models["setting"].get("system.baseUrl") || "localhost:3000";
+        const assignmentLink = `http://${baseUrl}/session/${studySession.hash}`;
+
+        // Get email content from template or fallback. Set context.link so ~link~ resolves to /session/ for the reviewer.
+        const emailContent = await getEmailContent(
+            "email.template.assignment",
+            "CARE - New Review Assignment",
+            `Hello,
```
dennis-zyska: Those default messages doesn't look good in the code, maybe we can create file on the disk that were we read from in case there is not template defined. Also for the other default emails like in the login, that way at least the code looks better.

backend/webserver/sockets/document.js
```diff
@@ -305,6 +307,35 @@ class DocumentSocket extends Socket {
...
+                    data.templateId,
+                    {
+                        baseUrl: baseUrl
+                    },
+                    this.models,
+                    options
+                );
+
+                // Overwrite the empty delta file with resolved content
+                const deltaFilePath = path.join(UPLOAD_PATH, `${doc.hash}.delta`);
```
dennis-zyska: Here as well, we need to make sure we don't forget the edits, delta could be overwritten at any time and only holds for basic documents not the study documents

backend/webserver/sockets/study.js
```diff
@@ -71,6 +72,119 @@ class StudySocket extends Socket {
...
+            for (const sessionOwnerId of userIds) {
+                try {
+                    const user = await this.models['user'].getById(sessionOwnerId);
+                    if (!user || !user.email) {
+                        this.logger.warn(`Cannot send study closed email: user ${sessionOwnerId} has no email`);
+                        continue;
+                    }
+
+                    const emailContent = await getEmailContent(
+                        "email.template.studyClosed",
```
dennis-zyska, line 110: Move to file as mentioned in a previous comment

backend/webserver/sockets/study_session.js
```diff
@@ -40,19 +41,191 @@ class StudySessionSocket extends Socket {
...
+        if (!user || !user.email) {
+            this.server.logger.warn(`Cannot send session start email: user ${study.userId} has no email`);
+            return;
+        }
+
+        const baseUrl = await this.models["setting"].get("system.baseUrl") || "localhost:3000";
+        const emailContent = await getEmailContent(
+            "email.template.sessionStart",
+            "CARE - Review Session Started",
+            `Hello,
```
dennis-zyska: Move content to disk file

backend/webserver/sockets/study_session.js
```diff
@@ -40,19 +41,191 @@ class StudySessionSocket extends Socket {
...
+
+        // Get baseUrl from settings
+        const baseUrl = await this.models["setting"].get("system.baseUrl") || "localhost:3000";
+        const reviewLink = `http://${baseUrl}/review/${studySession.hash}`;
+
+        // Get email content from template or fallback
+        const emailContent = await getEmailContent(
+            "email.template.sessionFinish",
+            "CARE - Review Session Completed",
+            `Hello,
```
dennis-zyska: move content to disk file

dennis-zyska (changes requested):
I reviewed the first half of the code, but I have an appointment now, so I will submit the review and come back to it later.

backend/webserver/sockets/template.js
```diff
@@ -0,0 +1,778 @@
...
+      draft: true,
+      order: idx + 1,
+      ...op,
+    }));
+
+    await this.models["template_edit"].bulkCreate(bulkEdits, {
+      transaction: options.transaction,
+    });
+
+    return { success: true };
```
dennis-zyska: We don't need to return this, it is already done by the createSocket class if we don't throw an error. You get this with the callback in the frontend (see other socket requests)

backend/webserver/sockets/template.js
```diff
@@ -0,0 +1,778 @@
...
+   * @param {number} data.id               Template ID to update (required)
+   * @param {string} [data.name]           New name
+   * @param {string} [data.description]    New description
+   * @param {string} [data.defaultLanguage] Default language code
+   * @param {boolean} [data.published]     New published flag (can only be set to true, cannot be unpublished)
+   * @param {Object} options
+   * @param {Object} options.transaction
+   * @returns {Promise<Object>}
+   */
+  async updateTemplate(data, options) {
```
dennis-zyska: This function doesn't have any special requirements that can't be handled via a beforeUpdate hook or is not already implemented in our updateData functionality (https://github.com/UKPLab/CARE/blob/main/backend/webserver/sockets/app.js#L64) right? So we should add no functions that are not absolutely necessary to reduce code and maintainability. Please also check the other functions in this class.

backend/webserver/sockets/template.js
```diff
@@ -0,0 +1,778 @@
...
+   * Get list of language codes that have content for a template
+   *
+   * @socketEvent templateGetLanguages
+   * @param {Object} data                  The data object
+   * @param {number} data.templateId       Template ID (required)
+   * @param {Object} options
+   * @param {Object} options.transaction
+   * @returns {Promise<Array<string>>}
+   */
+  async getLanguages(data, options) {
```
dennis-zyska, line 297: if we retrieve all templates in the frontend, aren't there also other languages retrieved or only the language that is currently selected? because otherwise we already have the languages in the frontend right?

backend/webserver/sockets/template.js
```diff
@@ -0,0 +1,778 @@
...
+   * @socketEvent templateAddLanguageContent
+   * @param {Object} data                  The data object
+   * @param {number} data.templateId       Template ID (required)
+   * @param {string} data.language         Language code (required)
+   * @param {Object} [data.content]        Content to copy (optional; if omitted, creates minimal empty content)
+   * @param {Object} options
+   * @param {Object} options.transaction
+   * @returns {Promise<Object>}
+   */
+  async addLanguageContent(data, options) {
```
dennis-zyska: Isn't it just `addContent` and language is just another variable that is defined in addition?

backend/db/migrations/20251210121131-create-template.js
```diff
@@ -0,0 +1,75 @@
...
+      userId: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+        references: {
+          model: 'user',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+      },
+      published: {
```
dennis-zyska: Is there a reason why it is published and not public?

backend/webserver/Socket.js
```diff
@@ -441,13 +441,22 @@ module.exports = class Socket {
         const filteredAccessMap = await this.filterAccessMap(accessMap, userId, rolesUpdatedAt);
         const relevantAccessMap = filteredAccessMap.filter(item => item.hasAccess);
         const accessRights = relevantAccessMap.map(item => item.access);
-        if (await this.isAdmin(userId, rolesUpdatedAt) || this.models[tableName].publicTable) { // is allowed to see everything
+        // Special handling for templates: even admins should see filtered results (own + published from others)
+        const isTemplateTable = tableName === "template";
+        if ((await this.isAdmin(userId, rolesUpdatedAt) || this.models[tableName].publicTable) && !isTemplateTable) { // is allowed to see everything (except templates)
```
dennis-zyska: I don't understand this exception? In general, we should not include any exceptions for specific tables here. If we need a new functionality here, we should implement that correctly and use static variables in the model class.

backend/webserver/Socket.js
```diff
@@ -441,13 +441,22 @@ module.exports = class Socket {
...
             // no adaption of the filter or attributes needed
         } else if (this.models[tableName].autoTable && 'userId' in this.models[tableName].getAttributes() && accessRights.length === 0) {
-            // is allowed to see only his data and possible if there is a public attribute
+            // is allowed to see only his data and possible if there is a public or published attribute
             const userFilter = {};
             if ("public" in this.models[tableName].getAttributes()) {
                 userFilter[Op.or] = [{userId: userId}, {public: true}];
+            } else if ("published" in this.models[tableName].getAttributes()) {
```
dennis-zyska: why not just using `public` as the column name?

backend/webserver/Socket.js
```diff
@@ -441,13 +441,22 @@ module.exports = class Socket {
...
+                if (tableName === "template" && typeof this.models[tableName].getUserFilter === "function") {
```
dennis-zyska: here as well, please don't use any exceptions here in that class. We use static variables in the model class for such functionalities, that allows us to control the behaviour also for other models if needed

backend/webserver/Socket.js
```diff
@@ -854,8 +869,13 @@ module.exports = class Socket {
                 this.io.to(socket.id).emit(tableName + "Refresh", data);
                 continue
             }
-            // if socket is admin or table is public, also just send
-            if (await this.isAdmin(userId, rolesUpdatedAt) || this.models[tableName].publicTable) {
+            const isTemplateTable = tableName === "template" &&
```
dennis-zyska: here as well, don't use specific table names in those general functions

backend/webserver/Socket.js
```diff
@@ -866,6 +886,20 @@ module.exports = class Socket {
                 continue;
             }
             allFilter = filtersAndAttributes.filter;
+            // For template table, non-admins must also receive updates to templates that are the source of their copies
+            if (isTemplateTable && !isAdmin && allFilter[Op.or]) {
```
dennis-zyska: and here as well

docs/source/for_developers/frontend/components/components.rst
```diff
@@ -141,6 +142,24 @@ By separating session resumption into its own component, CARE cleanly distinguis
 
 -----
 
+Templates
+---------
+
+The :doc:`Templates <templates>` component provides **email and document content templates** available from the Dashboard.  
+Templates are used for system emails (e.g. password reset, verification, registration), session start/finish and assignment emails, study-closed notifications, and pre-filled document content.
+
+- **Template types** 1–6: Email - General, Email - Study Session, Email - Assignment, Document - General, Document - Study, Email - Study Close.
```
dennis-zyska: It would be better to show it as a table or the type value direct by the definition

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
+Templates
+=========
+
+The **Templates** system provides email and document content templates that can be used for system emails, session and assignment notifications, study-closed emails, and pre-filled document content.  
+Templates are edited in the same Quill-based Editor as documents; placeholder resolution is done by the backend when the template is used.
+
+Key features include:
+
+  - **Template types** 1–6 (Email - General, Study Session, Assignment, Document - General, Document - Study, Email - Study Close), each with a fixed set of placeholders resolved at runtime.
```
dennis-zyska: here as well, provide a table or better align the type values with the definition, otherwise the reader needs to count themselve

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
...
+  - **Multi-language content** stored in ``template_language_content``; default language on the ``template`` row.
+  - **TemplateEditor** and **TemplateConfigurator** (Placeholders sidebar) shown when the Editor is opened with a template (``templateId`` provided).
+  - **Toolbar and editor behavior** controlled by the same settings as the document editor (see :ref:`Editor Settings <editor-settings-ref>`).
+
+Overview
+--------
+
+Templates are listed and created from **Dashboard → Templates** (nav from ``nav_element``; see migration ``extend-nav_element-templates``).
```
dennis-zyska: we don't need to reference the `nav_element` table here, that should be clear (better if you want to hint that to the user is to reference other parts of the documentation (like the navigation in that case))

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
...
+      <TemplateConfigurator />
+    </template>
+
+
+Template Types and Placeholders Resolved at Runtime
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+At resolution time, only the following placeholder keys are substituted (source: ``PLACEHOLDERS_BY_TYPE`` in ``backend/utils/templateResolver.js``).
+
+- **Type 1 (Email - General):** ``username``, ``firstName``, ``lastName``, ``link``
```
dennis-zyska: you can also reference for the type to that location

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
...
+At resolution time, only the following placeholder keys are substituted (source: ``PLACEHOLDERS_BY_TYPE`` in ``backend/utils/templateResolver.js``).
+
+- **Type 1 (Email - General):** ``username``, ``firstName``, ``lastName``, ``link``  
+- **Type 2 (Email - Study Session):** ``username``, ``link``  
+- **Type 3 (Email - Assignment):** ``username``, ``assignmentType``, ``assignmentName``, ``link``  
+- **Type 4 (Document - General):** none  
+- **Type 5 (Document - Study):** none  
+- **Type 6 (Email - Study Close):** ``username``, ``studyName``
+
+Where Each Type Is Used
```
dennis-zyska: why not having a table and merge both, type and usage explanation?

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
...
+- **template_language_content** — content (Quill Delta) per template and language.  
+- **template_edit** — draft edits per template and language.  
+- **template_placeholder_mapping** — placeholder keys and labels per template type (used by the frontend sidebar; resolution rules live in the resolver).
+
+Location: ``backend/utils/templateResolver.js``
+
+Placeholder resolution is implemented there: ``resolveTemplate`` (returns HTML for emails) and ``resolveTemplateToDelta`` (returns Delta for document creation).  
+Only placeholders listed in ``PLACEHOLDERS_BY_TYPE`` for the template's type are substituted at runtime.
+
+Implementing the Template Editor
```
dennis-zyska, line 37: what I'm missing here is how the edits will be merged in the content

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -0,0 +1,101 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Type 1** — Auth/system emails: settings ``email.template.passwordReset``, ``email.template.verification``, ``email.template.registration``. Location: ``auth.js``  
+- **Type 2** — Session start/finish emails: settings ``email.template.sessionStart``, ``email.template.sessionFinish``. Location: ``study_session.js``  
+- **Type 3** — Assignment emails: setting ``email.template.assignment``. Location: ``assignment.js``  
+- **Type 4** — Pre-fill document content when creating a document with ``templateId`` (``createDocument``). Location: ``document.js``  
+- **Type 5** — Document template for a study step (when creating document from template). Location: ``study_step.js``  
+- **Type 6** — Study-closed emails: setting ``email.template.studyClosed`` (``sendStudyClosedEmails``). Location: ``study.js``
+
+Adding a New Template Type or Placeholder
```
dennis-zyska: Actually, if I want to add a placeholder and I'm not experienced, I would miss some information. Maybe provide a specific example

docs/source/for_researchers/templates.rst
```diff
@@ -0,0 +1,90 @@
...
+Where Templates Are Used
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **System/auth emails** — Settings (password reset, verification, registration).  
+- **Session start and session finish emails** — Sent when a participant starts or finishes a session; template set in Settings; study must have **Enable Email Notifications** on.  
+- **Assignment emails** — Sent when a reviewer is assigned; template set in Settings; checkbox in Single/Bulk Assignment modal must be checked when creating the assignment.  
+- **Study closed emails** — Sent when a study is closed to participants with open sessions; template set in Settings; study must have **Enable Study Close Email Notifications** on.  
+- **Document creation** — When creating a document, you can select a document template (type 4 or 5) to pre-fill the content.
+
+Publishing and Deletion
```
dennis-zyska, line 91: It would be also nice to extend the documentation with the update mechanism

frontend/src/basic/navigation/Topbar.vue
```diff
@@ -201,7 +201,11 @@ export default {
       await this.$router.push("/login");
     },
     async toHome() {
-      await this.$router.push('/dashboard');
+      if (this.$route.path.startsWith('/template/')) {
```
dennis-zyska, line 204: what is the purpose of this?

frontend/src/basic/Form.vue
```diff
@@ -86,6 +86,96 @@
           />
         </span>
       </div>
+      
+      <!-- Collapsible Section -->
+      <div v-if="advancedFields.length > 0" class="col-12">
+        <Collapsible title="Advanced Settings" :collapsed="true">
+          <div class="row g-3">
+            <div
```
dennis-zyska: From here, there is a lof of multiple code, maybe we can add a new component and just include it here for the split between regular and advanced fields to reduce doubled code

frontend/src/basic/Table.vue
dennis-zyska, line 1: Please use in the future for such functionality and also the collapsing an additional pull request

frontend/src/components/dashboard/documents/CreateModal.vue
```diff
@@ -58,17 +74,29 @@ export default {
...
     };
   },
   computed: {
     selectedProjectId() {
       return this.$store.getters["settings/getValueAsInt"]("projects.default");
     },
+    documentTemplates() {
+      const currentUserId = this.$store.getters["auth/getUserId"];
+      // Own templates only (Type 4 Document - General), including copies
+      return this.$store.getters["table/template/getAll"]
```
dennis-zyska, line 88: How can we make sure the templates are loaded? Please add subscription to the table if it used here

frontend/src/components/dashboard/documents/CreateModal.vue
```diff
@@ -83,11 +111,18 @@ export default {
...
-      this.$socket.emit("documentCreate", {
+      const createData = {
         type: this.documentType, 
         name: this.name,
         projectId: this.selectedProjectId,
-      }, (res) => {
+      };
+
+      // Add templateId only if selected
+      if (this.templateId && this.templateId !== "") {
```
dennis-zyska: why not assign a 0 for "no template", that way it is always a number and those lines are not needed. In backend just don't use the template if templateId is 0

frontend/src/components/dashboard/settings/SettingItem.vue
```diff
@@ -63,9 +75,53 @@ export default {
       collapsed: true
     };
   },
+  computed: {
+    user() {
+      return this.$store.getters["auth/getUser"];
+    },
+    emailTemplates() {
```
dennis-zyska, line 83: here in that component you should also subscribe the templates

frontend/src/components/dashboard/settings/SettingItem.vue
```diff
@@ -63,9 +75,53 @@ export default {
...
   methods: {
     toggleCollapse() {
       this.collapsed = !this.collapsed;
+    },
+    isEmailTemplateSetting(setting) {
+      return setting.key && 
+             setting.key.startsWith("email.template.") && 
+             (setting.type === "number" || setting.type === "integer");
+    },
+    getFilteredEmailTemplates(setting) {
```
dennis-zyska, line 106: that are a lot of exception, we should come up with a better solutions, maybe a special type in the settings table? maybe not for now, but think about it in the other settings refactoring to delete this code in the future.

frontend/src/components/dashboard/study/BulkAssignmentModal.vue
```diff
@@ -449,7 +462,10 @@ export default {
       value: true
     }]
   },
-    "submission"
+    "submission",
+    {
```
dennis-zyska, line 466: here you have the subscription, whenever a specific table data is needed in a component, we should subscribe it.

frontend/src/components/dashboard/templates/PublicTemplatesModal.vue
```diff
@@ -0,0 +1,181 @@
...
+  },
+  computed: {
+    userId() {
+      return this.$store.getters["auth/getUserId"];
+    },
+    ownTemplates() {
+      return this.$store.getters["table/template/getAll"]
+        .filter(t => t.userId === this.userId && !t.deleted);
+    },
+    publicTemplates() {
```
dennis-zyska, line 72: table subscriptions are also missing here

frontend/src/components/dashboard/templates/PublishModal.vue
```diff
@@ -0,0 +1,147 @@
...
+    }
+  },
+  data() {
+    return {
+      id: 0,
+      success: false,
+    }
+  },
+  computed: {
+    template() {
```
dennis-zyska, line 81: here as well, maybe it not needed here, because the modal is in the template dashboard component, but we never know, better subscribe once again and don't miss information (we merge the subscriptions in the backend anyway)

frontend/src/components/dashboard/templates/PublishModal.vue
```diff
@@ -0,0 +1,147 @@
...
+        this.$socket.emit("stats", {
+          action: "openModalPublishTemplate",
+          data: {templateId: this.id}
+        });
+      }
+    },
+    publish() {
+      this.$socket.emit("templateUpdate", {
+        id: this.id,
+        published: true,
```
dennis-zyska: see other comments, why not using "public" as column name and use existing functionalities

frontend/src/components/dashboard/templates/TemplateModal.vue
```diff
@@ -0,0 +1,153 @@
...
+          const data = res.success && res.data ? res.data : {};
+          const languagesArray = Array.isArray(data) ? data : (data.languages || []);
+
+          this.filterDefaultLanguageOptions(languagesArray);
+
+          this.$nextTick(() => this.$refs.coordinator.open(id, defaultValues));
+        });
+      },
+
+      filterDefaultLanguageOptions(languagesWithContent) {
```
dennis-zyska: can we just use a computed for that method?

frontend/src/components/dashboard/templates/TemplateModal.vue
```diff
@@ -0,0 +1,153 @@
...
+        const fields = this.$store.getters["table/template/getFields"];
+        if (!fields) return;
+        const defaultLangField = fields.find((f) => f.key === "defaultLanguage");
+        if (!defaultLangField || !defaultLangField.options) return;
+        defaultLangField.options = languagesWithContent.length > 0
+          ? SUPPORTED_LANGUAGES.filter((opt) => languagesWithContent.includes(opt.value))
+          : [...SUPPORTED_LANGUAGES];
+      },
+
+      filterTypeOptionsInStore() {
```
dennis-zyska: also here, is a computed possible?

frontend/src/components/dashboard/templates/TemplateModal.vue
```diff
@@ -0,0 +1,153 @@
...
+          }
+
+        if (!isEdit) {
+          delete payload.id;
+          payload.defaultLanguage = payload.defaultLanguage;
+          // Set minimal content for new templates
+          payload.content = { ops: [{ insert: "\n" }] };
+        }
+  
+        const eventName = isEdit ? "templateUpdate" : "templateAdd";
```
dennis-zyska: see also my other comment in the backend socket, why not using appDataUpdate?

frontend/src/components/editor/sidebar/TemplateConfigurator.vue
dennis-zyska, line 1: The file name is misleading, maybe just `Placeholder` , also for later generalisation

frontend/src/components/editor/template/TemplateEditor.vue
dennis-zyska, line 1: What I was a bit confused, you basically created another Editor here, but then integraded it in the original Editor? I think there are two ways, either you make the selection one higher level in the component Template.vue where you used the Editor or even better you integrate it well in the existing editor to reduce duplicate code. I'm find with the first solution for now to get it merged, but please have a look in the Editor.vue component and check what changes are really necessary or which can be moved to Template.vue

dennis-zyska (changes requested):

frontend/src/basic/navigation/Topbar.vue
mohammadsherif0: To ensure back button from template editor returns to templates dashboard instead of documents dashboard

backend/webserver/sockets/template.js
mohammadsherif0: When we load templates in the frontend, we only retrieve the template rows, not their template_content rows. The per-language contents lives in the template_content table and are fetched separately, so we do not already have all languages in the frontend.

frontend/src/components/dashboard/settings/SettingItem.vue
mohammadsherif0: I will keep that in mind while working on the settings page changes.

backend/webserver/routes/auth.js
```diff
@@ -120,22 +115,19 @@ module.exports = function (server) {
...
+                context.options || {}
             );
-            
-            return {
-                subject: fallbackSubject, // Keep same subject for now (could be from template later)
-                body: resolvedHtml,
-                isHtml: true
-            };
+            const fallback = await getEmailFallbackContent(fallbackKey, context);
+            return { subject: fallback.subject, body: resolvedHtml, isHtml: true };
```
dennis-zyska, line 121: does the `resolvedHtml` must be fallback.body?

dennis-zyska (changes requested):
Thanks for the revision, looks good, I think there is just a minor bug, once fixed we can merge it.

backend/webserver/routes/auth.js
mohammadsherif0: The body here should stay as resolvedHtml. Since fallback.body is the plain text disk fallback for when no template is used. Where as this uses the DB template for the HTML body and only uses the fallback file for the subject.

dennis-zyska (approved):

PR #124, opened 2026-03-16
fix: ensure email verification setting is treated as a string in auth…

### LINKED ISSUE #121: [BUG] Verification Link Notifaction Confusing
# Describe the Bug:  

When email verification is turned off via the settings and you register a new account we still get a toast "verification link has been sent to your email".

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

See below, first response

# Expected Behavior:  

Not showing a notification.

# Actual Behavior:  

It shows a notification

# System Information: 

- **Device (OS/Type):** Ubuntu 20.x
- **Browser (+ Version):** Firefox
- **CARE Software (Version/Branch):** dev

# Additional Information:  

There is a setting for email notifications being required which you can turn on and off (https://github.com/UKPLab/CARE/blob/84c904cbfb1cc9b1011628933dd1f652d7fbf203/backend/db/migrations/20250911121017-basic-setting-mail_verification.js#L11). This setting, as usually, is loaded into the vuex store in the frontend and accessible through the backend. As an admin you can turn this on and off in the settings view of the dashboard. 

The issue occurs on the register vue (https://github.com/UKPLab/CARE/blob/84c904cbfb1cc9b1011628933dd1f652d7fbf203/frontend/src/auth/Register.vue) where a new user self-registers. This seemingly triggers this toast: https://github.com/UKPLab/CARE/blob/84c904cbfb1cc9b1011628933dd1f652d7fbf203/frontend/src/auth/Register.vue#L316 -- however there is already a save guard loading the setting.

Steps: 
1. try recreating the error -- it might not be possible since the saveguard seems solid
2. investigate causes why setting is not correctly loaded
3. fix

AyyanYe:
### Description
This PR resolves the issue where the "Verification link sent" toast notification was incorrectly displayed during user registration even when email verification was disabled. 

### Changes
- Modified `frontend/src/auth/Register.vue` to use a robust type-safe check for the `app.register.emailVerification` setting.
- Wrapped the configuration check in a `String()` cast to ensure that both boolean and string types from the backend are evaluated correctly against the strict equality operator.

### Root Cause Analysis
The frontend logic was performing a strict equality check (`=== "true"`). Depending on how the configuration was loaded or cached, the value could be returned as a boolean `true/false`, causing the check to fail silently and default to the "verification required" state.

### Verification Results
- **Environment:** GitHub Codespace / Docker
- **Test Case 1 (Disabled):** Set `app.register.emailVerification` to `false`. Registered a new user. **Result:** Success toast shown, no verification link toast.
- **Test Case 2 (Enabled):** Set `app.register.emailVerification` to `true`. Registered a new user. **Result:** Verification link toast shown as expected.

### Related Issues

Fixes #121

### DIFF
--- a/backend/webserver/routes/auth.js
+++ b/backend/webserver/routes/auth.js
@@ -92,7 +92,7 @@ module.exports = function (server) {
             }
             
             // Check if email verification is required and if user has verified their email
-            const emailVerificationEnabled = await server.db.models['setting'].get("app.register.emailVerification") === "true";
+            const emailVerificationEnabled = String(await server.db.models['setting'].get("app.register.emailVerification")) === "true";
             if (emailVerificationEnabled && !user.emailVerified) {
                 return res.status(401).json({
                     message: "Please verify your email address before logging in.",
@@ -212,7 +212,7 @@ module.exports = function (server) {
             transaction = await server.db.models['user'].sequelize.transaction();
             
             // Check if email verification is enabled
-            const emailVerificationEnabled = await server.db.models['setting'].get("app.register.emailVerification") === "true";
+            const emailVerificationEnabled = String(await server.db.models['setting'].get("app.register.emailVerification")) === "true";
             
             const userData = {
                 firstName: data.firstName,
@@ -407,7 +407,7 @@ The CARE Team`);
     server.app.get('/verify-email', async function (req, res) {
         const {token} = req.query;
               
-        if(await server.db.models['setting'].get("app.register.emailVerification") !== "true") {
+        if(String(await server.db.models['setting'].get("app.register.emailVerification")) !== "true") {
             return res.status(400).send({message:"Email verification is disabled."});
         }
         if (!token) {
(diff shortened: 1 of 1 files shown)


dennis-zyska (commented):
I was wondering what type this variable is, since all settings are saved in the database as strings. This should already be a string, so your code should not change anything. Did you test the solution? Does it fix the notification?

AyyanYe:
Great question, @dennis-zyska. Yes, I have fully tested the solution in a Docker environment, and it successfully fixes the notification behavior.

You're right that the DB stores it as a string, but I observed that by the time the configuration reaches window.config in the browser, the values are being treated as raw Booleans (e.g., true instead of "true"). This often happens during the JSON.parse step in the backend's /config.js route:

```javascript
res.send(`window.config = JSON.parse(${JSON.stringify(JSON.stringify(config))})`);
```
Because the frontend logic uses a strict equality check (=== "true"), it fails when it encounters a Boolean true. The String() cast ensures the check remains robust regardless of how the JSON is parsed or if the data type shifts during transit. Without this change, the check defaults to false, which triggers the 'verification required' toast by mistake.

dennis-zyska (approved):
I see, that makes sense, thank you for the clarification. I approved the PR and will merge it into dev. Thank you for your contribution!

PR #129, opened 2026-03-18
Fix #122 comment collapse

### LINKED ISSUE #122: [BUG] Comment collapsing error
# Describe the Bug:  

Comments that were once collapsed and that are then deleted will cause frontend errors and if other non-deleted comments exist will cause pdfs to not load anymore.

# Reproducibilty:  

- [ ] Yes 
- [ ] No 
- [ ] Sometimes 

(Describe as detailed as possible so others can reproduce the issue.) 

1. Step 1 
2. Step 2 
3. ... 

# Expected Behavior:  

(Describe what should happen under normal circumstances.) 

# Actual Behavior:  

(Describe what actually happens instead.) 

# Screenshots:  

(If applicable, upload screenshots of the issue to illustrate the problem.) 

# System Information: 

- **Device (OS/Type):** (e.g., Windows 10, macOS Monterey, Android 12) 
- **Browser (+ Version):** (e.g., Chrome 98, Firefox 105) 
- **CARE Software (Version/Branch):** (Specify the version and branch of the software where the issue was found.) 

# Additional Information:  

(Provide any additional information that could be useful, such as logs from the browser console, relevant files, or external dependencies.)

junaidferoz:
I investigated the issue and fixed the frontend crash in the PDF highlights flow.

## Root cause

1. When a comment was collapsed, we store a comment_state entry (state=1).
2. If that comment (or reply) was later deleted, the comment_state could still exist temporarily and reference a comment id that is no longer present in the frontend store.

PDFHighlights assumed that state.commentId always resolves to a valid comment and directly accessed:

`comment.annotationId`

This caused:

1. TypeError: Cannot read properties of undefined (reading 'annotationId')
2. Follow-up watcher/component update errors in PDFViewer
3. PDF view breaking after collapse + delete scenarios

## Fix implemented (frontend)
In `frontend/src/components/annotator/pdfViewer/Highlights.vue`:
1. Made` getAnnotationByCommentState(state)` null-safe:
-   return null if comment is missing
-   return null if annotation is missing
2. Added guard logic in the commentStates watcher:
-  skip processing when annotation is null
-  avoid unsafe property reads on missing objects

## Why this works
The component now safely ignores orphaned/stale comment_state references instead of throwing, so deleted collapsed comments no longer crash highlight/watcher updates and PDF rendering remains stable.


frontend/package-lock.json
dennis-zyska, line 1: Please make sure that this file is not part of the commit

logs/activity-2026-03-17.log
dennis-zyska, line 1: Please make sure that this file is not part of the commit

dennis-zyska (changes requested):
Hi, thanks for the fix. There are additional log files in the commit, that should not be on the branch. Please delete and check the .gitignore.

junaidferoz:
Hi @dennis-zyska,
I have made the changes as per your comment. Could you please check and let me know what to do next with regards to this issue.

dennis-zyska (approved):

PR #132, opened 2026-03-23
Fix ESLint configuration and resolve all unused variable errors

akash9676:
Removed all unused variables, imports, and parameters from the frontend.
### New User Features
- NA
### New Dev Features
- Added a clear convention: prefix unused catch/callback parameters with `_`
### Improvements
- Removed unused imports and variable declarations across 32 files
- Cleaner, more maintainable codebase with no dead code warnings
### Bug Fixes
- Fixed `make lint` crash due to outdated config format
### Known Limitations
- Other lint warnings (unrelated to unused variables) are not addressed in this MR
### Future Steps
- Clean up remaining lint warnings in follow-up MRs

### DIFF
--- a/frontend/src/basic/editor/JsonEditor.vue
+++ b/frontend/src/basic/editor/JsonEditor.vue
@@ -137,7 +137,7 @@ export default {
 
     try {
       window.removeEventListener("click", this.leaveEditModeListener);
-    } catch (e) {
+    } catch (_e) {
       // do nothing
     }
   },
@@ -152,7 +152,7 @@ export default {
       try {
         JSON.parse(this.contentText);
         return true;
-      } catch (e) {
+      } catch (_e) {
         return false;
       }
     },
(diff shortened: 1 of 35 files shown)


frontend/src/basic/editor/JsonEditor.vue
```diff
@@ -197,7 +197,7 @@ export default {
             message: "Json copied to clipboard!",
             variant: "success"
           });
-        } catch ($e) {
+        } catch (_$e) {
```
dennis-zyska: if we align it, should we do `_error` instead of `_$e` ?

frontend/src/components/dashboard/workflows/WorkflowCreateModal.vue
```diff
@@ -42,7 +42,7 @@ export default {
           message: "Workflow link copied to clipboard!",
           variant: "success"
         });
-      } catch ($e) {
+      } catch {
```
dennis-zyska: here you removed it completely, please align with the other cases

dennis-zyska (changes requested):
Error Variables are sometimes `_$e`, `_e` or `_error`, we should align them. As variable should be self explainable we should use `_error` for all cases

frontend/src/basic/editor/JsonEditor.vue
akash9676: "_" should be fine

dennis-zyska (approved):

PR #133, opened 2026-03-24
[FEATURE] Add logo color customization

akash9676:
This feature allows admins to change the logo color from the settings page and shows the updated logo across key parts of the platform.

### New User Features

- Admins can change the logo color in Settings.
- A preview shows how the logo looks before saving.
- The updated logo appears on login, registration, reset password, and top navigation.

### New Dev Features

- Added a reusable logo component.
- Added a stored setting for logo color.

### Improvements

- Branding is now easier to manage.
- Logo text color adjusts for better visibility.

### Bug Fixes

- None

### Known Limitations

- Only the logo background color can be changed.
- Other logo parts are still fixed.

### Future Steps

- If any one want there own version of care instance with there own naming for this we need to add sub text in logo to distinguish between multiple instances like (test, EiWA etc..)


frontend/src/components/dashboard/Settings.vue
```diff
@@ -65,12 +65,18 @@
...
 
         <div v-else class="mt-3">
-          <SettingItem
-              v-for="(group, key) in groupedSettings"
-              :key="key"
-              :group="group"
-              :title="key"
-          />
+          <template v-for="(group, key) in groupedSettings" :key="key">
+            <LogoSettings
```
dennis-zyska: This should be moved into SettingItem (here: https://github.com/UKPLab/CARE/blob/dev/frontend/src/components/dashboard/settings/SettingItem.vue#L30) , we can set a special kind of type (e.g. `color`) here, where the color picker is selected. We should generalize it directly, in case we use it later for other parts

backend/db/migrations/20260209172847-basic-logo-settings.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+
+const settings = [
+  {
+    key: "logo.reBgColor",
+    value: "#ffe599",
+    type: "string",
```
dennis-zyska: use a type `color` instead of `string`, see my other comment too

dennis-zyska (changes requested):
Nice implementation, just a minor comment

dennis-zyska (approved):

PR #142, opened 2026-03-31
Feat 120 assigment dashboard

karimouf:
### Main Description

Implemented an Assignment Dashboard flow that shifts submission handling from admin-only/manual import to assignment-based self-service uploads, enabling both admins and users to manage submissions and revision workflows.

---

### User Features

* Create assignments and assign to roles
* Edit / Copy / Delete / Close assignments
* View submissions per assignment
* Submit multiple revisions
* Replace/Delete submissions (if not used in a study)
* View and manage assignment-bound submissions
* Submission metadata support (`description`)
* Study usage tracking via `studyUsageCount`
* Assignment-specific views for submissions
* Max revisions slider (including unlimited option)

---

### System Features

* Assignment Data Model

  * `title`, `description`, `public`
  * `studyId`, `workflowId`, `userId`
  * `start`, `end`, `closed`
  * `validationConfigurationId`
  * `assignedRoleIds`
  * `parentAssignmentId`, `previousSubmissionAssignmentId`
  * `allowReUpload`, `maxRevisions`
  * `deleted`, `deletedAt`, `createdAt`, `updatedAt`

* Submission Model Updates

  * `name`, `description` (metadata)
  * `assignmentId` (linkage)

* Access Control

  * Own + public visibility
  * Admin: full access
  * Owner-only actions (Edit/Delete/Share)

* Submission Pipeline

  * Assignment-based uploads
  * Revision chain enforcement

* Validation & Rules

  * Validation before persistence
  * Upload only when open (admin override)
  * Re-upload controlled via configuration

* Revision Control

  * Max revision limits
  * Standard naming (`revision X`)

* Study Protection

  * Prevents unsafe replace/delete using `studyUsageCount`

---

### Known Limitations

* User assignment view still missing some features (pending alignment/decision)


backend/db/migrations/20260316122741-create-assignment.js
```diff
@@ -0,0 +1,137 @@
...
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.createTable('assignment', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      title: {
```
dennis-zyska: we usually use `name` as the column name

backend/db/migrations/20260316122741-create-assignment.js
```diff
@@ -0,0 +1,137 @@
...
+        type: Sequelize.INTEGER,
+        allowNull: true,
+        references: {
+          model: 'project',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
+      },
+      studyId: {
```
dennis-zyska: Didn't we decide to remove this here?

backend/db/migrations/20260316122741-create-assignment.js
```diff
@@ -0,0 +1,137 @@
...
+        type: Sequelize.INTEGER,
+        allowNull: true,
+        references: {
+          model: 'study',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
+      },
+      workflowId: {
```
dennis-zyska: this as well?

backend/db/migrations/20260329120000-extend-assignment-assigned_role_ids.js
dennis-zyska, line 1: Can't we add this in the assignment creation migration? As long it is not merged to dev, I think it would be fine to do so. I'm also thinking, maybe it directly make sense here to adapt it also with an additional table instead of an array. Arrays in db tables are always a bad idea (also see https://en.wikipedia.org/wiki/Database_normalization)

backend/db/models/assignment.js
```diff
@@ -0,0 +1,207 @@
...
+		static accessMap = [
+			{
+				right: "frontend.dashboard.assignments.viewAll",
+				columns: this.getAttributes(),
+			},
+		];
+		/**
+		 * Apply visibility filter for assignments based on assigned roles.
+		 * Non-admin users can always see their own assignments and assignments
+		 * that are assigned to at least one of their roles.
```
dennis-zyska: Please add parameter information (userId + return)

backend/db/models/assignment.js
```diff
@@ -0,0 +1,207 @@
...
+			}
+
+			return {
+				[Op.or]: [
+					{ userId },
+					{ assignedRoleIds: { [Op.overlap]: roleIds } },
+				],
+			};
+		}
+		static fields = [
```
dennis-zyska, line 14: move this before the function declarations

backend/db/models/assignment.js
```diff
@@ -0,0 +1,207 @@
...
+			},
+			{
+				key: "description",
+				label: "Description:",
+				help: "Optional description shown for this assignment.",
+				type: "textarea",
+				required: false,
+			},
+			{
+				key: "studyId",
```
dennis-zyska: remove this

backend/db/models/assignment.js
```diff
@@ -0,0 +1,207 @@
...
+				options: {
+					table: "study",
+					name: "name",
+					value: "id",
+				},
+				required: false,
+				help: "Select a study assignment source.",
+			},
+			{
+				key: "workflowId",
```
dennis-zyska: remove this

backend/db/models/assignment.js
```diff
@@ -0,0 +1,207 @@
...
+			assignedRoleIds: {
+				type: DataTypes.ARRAY(DataTypes.INTEGER),
+				allowNull: true,
+				defaultValue: [],
+			},
+			parentAssignmentId: DataTypes.INTEGER,
+			allowReUpload: DataTypes.BOOLEAN,
+			notifyOnSubmissionUpload: {
+				type: DataTypes.BOOLEAN,
+				allowNull: false,
```
dennis-zyska: are those information needed to define here again? I think it is already defined by the migration/db right?

backend/db/models/document.js
```diff
@@ -52,6 +52,13 @@ module.exports = (sequelize, DataTypes) => {
                 required: false,
                 default: false
             },
+            {
```
dennis-zyska: I don't think this make sense here/is needed, the user should not edit it, it should be set only by the hooks in the study

backend/db/models/document.js
```diff
@@ -330,6 +337,11 @@ module.exports = (sequelize, DataTypes) => {
             projectId: DataTypes.INTEGER,
             submissionId: DataTypes.INTEGER,
             originalFilename: DataTypes.STRING,
+            studyUsageCount: {
+                type: DataTypes.INTEGER,
+                allowNull: false,
```
dennis-zyska: here as well, I think this is already defined by the database itself

backend/db/models/study.js
```diff
@@ -297,6 +336,16 @@ module.exports = (sequelize, DataTypes) => {
...
             }
 
+            const usedDocumentIds = [...new Set(
+                Object.values(studyStepsMap)
+                    .map((step) => step.documentId)
+                    .filter(Boolean)
+            )];
+
+            for (const documentId of usedDocumentIds) {
+                await Study.updateDocumentStudyUsageCount(documentId, options);
```
dennis-zyska: I think this should be called by the hook, or maybe it would be better to place it in the study_step.js model. That feels a bit cleaner.
What happens if the study is deleted? Should we update the document usage count as well? I guess it depends on whether we show this value in the frontend, so there is no mismatch.
But since the annotations are still available, we probably should not show it, because we cannot delete it completely. 
TBD

backend/db/models/template.js
```diff
@@ -66,7 +66,7 @@ module.exports = (sequelize, DataTypes) => {
         /**
          * Override getAutoTable to apply custom filtering for templates:
          * - All users (including admins): own templates OR public templates from others
-         * - Non-admins: exclude email templates (types 1, 2, 3, 6) - admin-only
+         * - Non-admins: exclude email templates (types 1, 2, 3, 6, 7) - admin-only
```
dennis-zyska, line 69: as there is a new type, we should also update the documentation accordingly

backend/webserver/sockets/document.js
```diff
@@ -981,12 +1166,152 @@ class DocumentSocket extends Socket {
...
             this.logger.error(error);
             throw new Error(error);
         }
     }
 
+    /**
+     * Replace an existing assignment submission by creating a new one,
+     * deleting the old one, and reconnecting submission chain pointers.
+     *
+     * @param {Object} data
```
dennis-zyska: The descriptions for the variables are missing

backend/webserver/sockets/document.js
```diff
@@ -946,28 +1034,125 @@ class DocumentSocket extends Socket {
...
+                    where: {
+                        assignmentId,
+                        userId,
+                        deleted: false,
+                    },
+                    raw: true,
+                    transaction,
+                });
+
+                const submissionById = new Map(assignmentSubmissions.map((submission) => [submission.id, submission]));
```
dennis-zyska, line 1129: This feels more complex than necessary. We are loading all submissions and reconstructing the chain in application code just to determine the latest submission and enforce maxRevisions. If there is only one submission chain per user/assignment, could we simplify this to fetch the latest submission directly and use a count query instead? Also, resolveChainTailId appears to be unused.

```
const submissionCount = await this.models["submission"].count({
        where: {
            assignmentId,
            userId,
            deleted: false,
        },
        transaction,
    });
```

backend/webserver/sockets/document.js
```diff
@@ -981,12 +1166,152 @@ class DocumentSocket extends Socket {
...
+            throw new Error(`Submission with id ${submissionId} not found for this assignment`);
+        }
+
+        const isOwner = this.userId === oldSubmission.userId;
+        const hasRight = await this.hasAccess('frontend.dashboard.assignments.replaceDeleteSubmissions');
+        if (!isOwner && !hasRight) {
+            throw new Error("You are not allowed to replace this submission.");
+        }
+
+        const oldSubmissionDocuments = await this.models["document"].findAll({
```
dennis-zyska, line 1271: We replacing a submission with a specific id here, right? Do we need .findAll and .some then? That also needs more compute than just retrieve by the id directly

backend/webserver/sockets/document.js
dennis-zyska, line 1: it feels like a lot of commands should be move to the model files and just called here, it would be much cleaner to read as the most lines are just database stuff. Permissions checks should be here, but database stuff should be in the model files

backend/webserver/Socket.js
```diff
@@ -643,7 +677,7 @@ module.exports = class Socket {
 
         allFilter = filtersAndAttributes.filter;
         allAttributes = filtersAndAttributes.attributes;
-
+        console.log("Final filter for " + tableName + ": ", allFilter, allAttributes);
```
dennis-zyska: don't forget to remove console logs

frontend/src/components/dashboard/assignments/AdminAssignmentsView.vue
```diff
@@ -0,0 +1,446 @@
...
+        maxRevisions: assignment.maxRevisions ?? 1,
+        public: Boolean(assignment.public),
+        start: assignment.start ? new Date(assignment.start).toLocaleString() : "-",
+        end: assignment.end ? new Date(assignment.end).toLocaleString() : "-",
+        allowReUpload: Boolean(assignment.allowReUpload),
+      }));
+    },
+  },
+  methods: {
+    isAssignmentOwner(assignment) {
```
dennis-zyska: I think we can directly do it in the compute. I know it is used twice, but I think here it is more important to directly be able to see what is compared

frontend/src/components/dashboard/assignments/AdminAssignmentsView.vue
dennis-zyska, line 1: we should not call it `AdminAssignmentsView.vue`, better would be to have right, that we give admins to use this. In that way we can in the future in principle give this also another role and keep this more general

frontend/src/components/dashboard/assignments/AssignmentSubmissionsModal.vue
```diff
@@ -0,0 +1,197 @@
...
+      </span>
+    </template>
+  </BasicModal>
+  <ImportModal ref="importModal" />
+  <AssignmentUploadModal ref="uploadModal" />
+  <ConfirmModal ref="deleteConf" />
+</template>
+
+<script>
+import BasicModal from "@/basic/Modal.vue";
```
dennis-zyska, line 48: Don't forget to add docstrings also in the components that describes what the component should do, please also for the others

frontend/src/components/dashboard/Assignments.vue
```diff
@@ -0,0 +1,462 @@
...
+      }
+
+      return this.$store.getters["table/assignment/getFiltered"](
+        (assignment) => assignment.userId === this.userId || Boolean(assignment.public) 
+      ) || [];
+    },
+    assignmentTable() {
+      return this.assignments.map((assignment) => ({
+        ...assignment,
+        isOwner: this.isAssignmentOwner(assignment),
```
dennis-zyska: here I would also remove the function and show it directly, because of the clearness

frontend/src/components/dashboard/Submissions.vue
```diff
@@ -1,387 +1,224 @@
 <template>
-  <Card title="Submissions">
-    <template #headerElements>
-      <div class="btn-group gap-2">
-        <BasicButton
```
dennis-zyska, line 5: Where are all the remove buttons and functionalities? I didn't find it, in which dashboard?

dennis-zyska (changes requested):
Thank you, nice and fast implementation, I made some comments for improvements. Note that I didn't run the branch yet, so the feedback is only on the code. There are also some merge conflicts.

backend/db/models/template.js
karimouf: @mohammadsherif0

frontend/src/components/dashboard/Assignments.vue
karimouf: wdym by show it directly? i feel it's clear enough

frontend/src/components/dashboard/Submissions.vue
karimouf: which butotn are you refferring too

backend/webserver/sockets/document.js
karimouf: are you reffering to my added functions or in general?

backend/db/models/study.js
karimouf: well I know the annotations are still available but we count how many studies the document is used in so it would only make sense to update it when a study is deleted even if it still has undeleted data related to it (correct if I'm wrong)

backend/webserver/sockets/document.js
karimouf: well my idea was to enable the replacement of all submission doesn't necessary have to be the latest one

backend/webserver/sockets/document.js
karimouf: well we find all documents under that submission so yes I need to find all in that case (in case of eiwa we have pdf and zip files)

frontend/src/components/dashboard/Assignments.vue
dennis-zyska: instead of `isOwner: this.isAssignmentOwner(assignment)` write `isOwner: assignment.userId === this.userId`

dennis-zyska: in that way you can directly see, the userId in assignment variable is compared to this.userId

frontend/src/components/dashboard/Submissions.vue
dennis-zyska: For `Assign Group`, `Publish Reviews`, `Publish Assessment`, ... in which component are they now?

backend/db/migrations/20260418000000-create-assignment_share.js
dennis-zyska, line 1: lets rename it to `assignment_share` to align it with the other tables for sharing

backend/db/models/document.js
dennis-zyska: Did you cleaned the db model files to make it more readable?

backend/webserver/sockets/document.js
dennis-zyska: both

dennis-zyska (changes requested):

frontend/src/components/dashboard/Submissions.vue
karimouf: assign group no longer applicable other function are moved to either submission ot studies

dennis-zyska (approved):

Issue #68, opened 2026-01-21
[FEATURE] Extract submissions


melolw:
## Is your feature request related to a problem? Please describe 

The feature is part of the broader goal to be able to extract the dataset (or part of it). 

## Describe the solution you'd like 

When clicking the 'download' button featured on the project dashboard, we should be able to select 'submissions' and download the submissions as a zip containing for each submission both pdf and zip. There should be options to select which submissions (draft or final?) and a confirmation step with the number of submissions that will be downloaded.

## Describe alternatives you've considered 

Currently the only alternative is extracting the submissions from Moodle.

## Additional context 

//

PR #146, opened 2026-04-07
Issue #68: Extract submissions Feature

melolw:
This feature introduces a way to download all submissions from a project using stream download. 
By going to Projects -> Export data on one project, you're able to choose "Extract submissions" in the dropdown, activating new steps to personalize your download. First selecting students, then creating aliases if necessary and finally downloading the data into a zip directly on your end device. 

Depending on the current user, the naming changes: if the user is an admin, the full names are used. Else, the username. Those are overriden if anonymization is selected, because then the aliases are used.

Limitations:
- Naming of files are done with regex saved in the configurations. If no standard names are mentionned in the configuration, the name will be the hash of the document. In the future, we want to save the standard name chosen in the configuration to be saved as the name of the file.
- Anonymization in currently handled only for structures with latex, need to check and replace name for all text files in the future so that word file for example are also handled.


backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+
+        // Auth checking
+        const currentUserId = req.user?.id;
+        if (!currentUserId) return res.status(401).send("Log in required");
+
+        const currentUser = await server.db.models.user.findByPk(currentUserId);
+        if (!currentUser) return res.status(401).send("User not found");
+
+        // check if user has right to see full names
+        let hasPrivateInfoRight = false;
```
dennis-zyska, line 21: we should use here an existing function hasAccess(): https://github.com/UKPLab/CARE/blob/dev/backend/webserver/Socket.js#L247 

In that way, in case of changes, we need to do it only once

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
+const archiver = require('archiver');
+const path = require('path');
+const fs = require('fs');
+const { faker } = require('@faker-js/faker');
+const crypto = require('crypto');
+const JSZip = require('jszip');
+
+module.exports = function (server) {
+
+    server.app.post('/export/project/stream', async function (req, res) {
```
dennis-zyska: Let's change the path to `/export/stream`, we should reuse this endpoint and handle multiple cases here. Then we also don't need to write the Auth checking code multiple times

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+    // HELPER FUNCTIONS
+
+    /**
+     * Generates a deterministic 32-bit unsigned integer seed by hashing a 
+     * provided seed integer with a user-specific salt using SHA-256.
+     * * @param {number} seedInt - The base integer seed (from the form input).
+     * @param {string} saltHex - The hex-encoded salt string from the user's database record.
+     * @returns {number} - A deterministic 32-bit unsigned integer (uint32) to be used as a seed for the Faker API.
+     */
+    function deriveUserSeed(seedInt, saltHex) {
```
dennis-zyska: This helper function can be moved to https://github.com/UKPLab/CARE/blob/dev/backend/webserver/auth/utils.js

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * Opens a zip file, replaces the student's real name with a fake name in all .tex files,
+     * and returns the modified zip as a Buffer.
+     * * @param {string} filePath - Path to the original zip file on disk
+     * @param {string} realName - The student's real name to search for
+     * @param {string} fakeName - The anonymized name to insert
+     * @returns {Promise<Buffer>} - The newly generated zip file buffer
+     */
+    async function anonymizeZipFile(filePath, realName, fakeName) {
```
dennis-zyska: Better use a function name that describes better what the function does, e.g. replaceAuthorInZip

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+     * * @param {Array<Object>} users - Array of user objects from the database.
+     * @param {boolean} shouldAnonymize - Whether the export should use fake names.
+     * @param {boolean} hasPrivateInfoRight - Whether the current user is allowed to see/export full names.
+     * @param {number|string} fakerSeed - The base integer seed (from the form input).
+     * @param {string} salt - The hex-encoded salt string from the user's database record.
+     * @returns {Object} An object containing:
+     * - userMapping: An object mapping user IDs to their generated anonymized names.
+     * - mappingCsv: A CSV-formatted string containing the mapping (conditionally includes real names).
+     */
+    function buildUserMapping(users, shouldAnonymize, hasPrivateInfoRight, fakerSeed, salt) {
```
dennis-zyska: In frontend we use papaparse for it: https://github.com/UKPLab/CARE/blob/dev/frontend/src/assets/utils.js#L49, maybe we can also use it here instaed of builden the CSV by our own

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+            });
+        }
+        return { userMapping, mappingCsv };
+    }
+
+    /**
+     * Does the fetching, filtering, and archiving of student submissions for a specific project.
+     * Handles file renaming based on validation rules and manages directory structures
+     * (Student Name/Version/File) within the ZIP archive.
+     * * @param {Object} server - The server instance providing database models and Sequelize operators.
```
dennis-zyska: Is there a reason why in all docstring there are two `* *` before each first @param

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+     * @returns {Promise<void>} - Resolves once all submissions have been processed and added to the archive.
+     */
+    async function processSubmissionsExport(server, projectId, userIds, users, shouldAnonymize, hasPrivateInfoRight, userMapping, baseFolderName, archive) {
+        
+        // Fetch all top-level submissions for the selected users
+        const submissions = await server.db.models.submission.findAll({
+            where: {
+                projectId,
+                userId: userIds,
+                parentSubmissionId: null // Using null directly instead of Op.is
```
dennis-zyska: we should also filter for delete:false (usually we try to use functions from here https://github.com/UKPLab/CARE/blob/dev/backend/db/MetaModel.js), but in your case it is fine to directly use findAll, but the frontend is filtered for delete:false, so we also need to do it here to be consistent

backend/webserver/routes/export.js
```diff
@@ -0,0 +1,315 @@
...
+            },
+            include: [{
+                model: server.db.models.document,
+                as: 'documents'
+            }]
+        });
+
+        // Fetch configurations for file renaming rules
```
dennis-zyska: please make clear in the comment, that those are the validation configuration for the document, with what the documents are validated on upload, otherwise it could be misleading

frontend/src/components/dashboard/projects/export/StepConfirmDownload.vue
```diff
@@ -0,0 +1,65 @@
...
+            {{ row.studentName || row.userName }} ({{ row.fileCount }} files)
+          </li>
+        </ul>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import BasicLoading from "@/basic/Loading.vue";
```
dennis-zyska, line 40: Also frontend files should have a docsctring, like here: https://github.com/UKPLab/CARE/blob/dev/frontend/src/components/dashboard/Documents.vue#L56

Also for all further components

frontend/src/components/dashboard/projects/export/StepOptions.vue
```diff
@@ -0,0 +1,62 @@
...
+        id="fakerSeedInput"
+        v-model.number="seed" 
+        type="number" 
+        class="form-control form-control-sm" 
+        max="999999999"
+        style="max-width: 200px;"
+        placeholder="e.g. 846569412"
+      >
+      <small class="text-muted d-block mt-1">
+        Using the same seed generates the same aliases across different exports.
```
dennis-zyska: we should also mention somewhere, that it is combined with a individual user seed and the export from two users is therefore different

dennis-zyska (changes requested):
Nice implementation, thanks, there are just some minor comments for improving the code quality. Please also have a look for the merge conflicts.

backend/webserver/routes/export.js
melolw: since hasAccess() uses this.userInfo and we don't use socket connection in export.js, i don't think that's going to work as is. we could move the logic inside hasAccess() somewhere else where both Socket.js and export.js could use it?

backend/webserver/routes/export.js
melolw: done

backend/webserver/routes/export.js
melolw: maybe in auth/utils.js as well??

backend/webserver/routes/export.js
melolw: done 👍🏻

backend/webserver/routes/export.js
melolw: i added papaparse in the backend and used it for this function, i didn't have to use as much if-statements 🙂

backend/webserver/routes/export.js
melolw: no, i think it's a typo I made in the first one which i copy pasted over and over 😅

melolw: i edited all of them to remove it

backend/webserver/routes/export.js
melolw: done

backend/webserver/routes/export.js
melolw: changed the comment, hope it's more comprehensible

frontend/src/components/dashboard/projects/export/StepConfirmDownload.vue
melolw: done for all components

frontend/src/components/dashboard/projects/export/StepOptions.vue
melolw: done

backend/webserver/routes/export.js
melolw: you meant deleted right? not delete?

backend/webserver/routes/export.js
melolw: done, moved the const crypto = require('crypto'); as well to auth/utils.js

backend/webserver/routes/export.js
dennis-zyska: yes true, we can't use it directly, I think it is fine for now

dennis-zyska (approved):

PR #152, opened 2026-04-11
Feat 145 assignments email notification

mohammadsherif0:
### Main description

Notify the assignment owner by email when a student uploads or re-uploads a submission for an assignment. Admins can configure an HTML email template (type Email – Submission upload) via Settings, or rely on the disk fallback. Owners can turn notifications off per assignment with Email owner on submission upload.

### New user features

- Email to the assignment owner on first submission upload and on re-upload (when enabled).
- Assignment setting: Notify on Submission Upload.
- Optional admin-configured submission upload email template in Settings; otherwise default text from the fallback file.


karimouf (approved):
I think its good to go, I just have one comment I feel like adding a new email template is a lot of work it should be easier. there is a lot of changes just for adding an extra template

PR #158, opened 2026-04-14
[FEATURE] Integrated LiteLLM RPC

### LINKED ISSUE #138: [FEATURE] LiteLLM RPC Integration
### Summary
Add a new RPC service in the CARE backend to route LLM requests through LiteLLM instead of BrokerIO, reducing latency and complexity for general-purpose LLM calls.

### Motivation
Currently, all model requests including simple LLM prompts are routed through BrokerIO. BrokerIO was designed for orchestrating specialized NLP models and introduces unnecessary overhead (queuing, skill lookup, resource binding) for straightforward LLM API calls. A dedicated LiteLLM RPC path provides a lighter, faster route.

### Requirements
- Create a new `LiteLLMRPC` service extending the existing RPC base class
- Support communication with LiteLLM via WebSocket (consistent with CARE's RPC pattern)
- Route requests to external LLM APIs (OpenAI, Anthropic, etc.) and locally hosted models (Ollama) through LiteLLM
- Keep BrokerIO intact for specialized NLP skills
- The NLP service should determine the correct path (LiteLLM or BrokerIO) based on the request type
- Log all requests and responses in the database for cost tracking and research

### Acceptance Criteria
- [ ] A request from the frontend can reach an external LLM (e.g. GPT-4o) via LiteLLM and return a response
- [ ] A request from the frontend can reach a local model (e.g. Ollama) via LiteLLM and return a response
- [ ] Existing BrokerIO-based NLP skills continue to work without changes
- [ ] All LLM inputs and outputs are recorded in the database

### References
- Existing RPC base class: `backend/webserver/RPC.js`
- NLP service: `backend/webserver/services/nlp.js`
- LiteLLM docs: https://docs.litellm.ai/

akash9676:
This feature adds a new AI request path that lets the platform route chat-generation requests through LiteLLM, making it easier to use different model providers through one consistent flow.

### New User Features

- Send AI chat requests through a new unified LLM route.
- Receive clearer response and error feedback when AI service is unavailable.
- Check whether the LLM service is online before making requests.

### New Dev Features

- New LiteLLM RPC service added to local Docker setup.

### Improvements

- Existing NLP service now supports both skill requests and LLM requests in one place.
- Better service-status handling for LLM connectivity checks.
- More reliable local startup flow with LiteLLM included in `make docker`.

### Bug Fixes

- None

### Known Limitations

- Provider access (such as API keys/model access) must still be configured externally.
- No dedicated UI flow included in this change.

### Future Steps

- Creating a dashboard where a user or an admin can manage the LLM models


backend/webserver/rpcs/liteLLMRPC.js
```diff
@@ -0,0 +1,51 @@
+const RPC = require("../RPC.js");
+
+/**
+ * LiteLLMRPC - Routes LLM requests through LiteLLM for external and local model access
+ *
+ * Pure passthrough: the caller supplies the model, messages, API key, and any
+ * provider-specific parameters. Nothing is hardcoded here; the bridge forwards
+ * everything to LiteLLM as-is.
+ *
```
dennis-zyska: You can add yourself as an author here

backend/webserver/services/nlp.js
```diff
@@ -358,6 +369,80 @@ module.exports = class NLPService extends Service {
...
+    }
+
+    /**
+     * Return LiteLLM RPC status to the client.
+     * @param client
+     */
+    async #handleLLMGetStatus(client) {
+        const rpc = this.#getLiteLLMRPC();
+        if (!rpc) {
+            await this.send(client, "llmStatus", {online: false, error: "LiteLLM RPC not registered"});
```
dennis-zyska: I think we should not integrate it here like this. I think this.send() sends the data to the broker, we should create another new service, like `AI.js`. This send is old and does not use callbacks as we do between the frontend and backend.

https://python-socketio.readthedocs.io/en/latest/client.html#event-callbacks

For RPC we already do it like here: https://github.com/UKPLab/CARE/blob/dev/backend/webserver/RPC.js#L170

The basis for the new `AI.js` could be the `backgroundTask.js`
https://github.com/UKPLab/CARE/blob/8e6968d7d499d15352b40896e2a06116800d3122/backend/webserver/services/backgroundTask.js

In the frontend, there should be another file (e.g., component) for communicating with it. A complete vue component may be not the correct way, as we don't want to always show something in frontend, but use AI requests easily. Maybe we can use a plugin like this (https://github.com/UKPLab/CARE/blob/8e6968d7d499d15352b40896e2a06116800d3122/frontend/src/plugins/subscribeTable.js) to make the service available in each component or you may find a better way?

utils/rpcs/litellm/main.py
```diff
@@ -0,0 +1,84 @@
...
+    logger.setLevel(logging.INFO)
+
+    sio = socketio.Server(async_mode='threading', ping_timeout=120, ping_interval=25)
+
+    @sio.event
+    def connect(sid, environ, auth):
+        logger.info(f"Connection established with {sid}")
+
+    @sio.on("call")
+    def call(sid, data):
```
dennis-zyska: This is a health check, why not changing `call` to something like `healthy`, and add this to the Service.js as standard way. If not implement by the rpc, we can throw in backend "Not Implemented" error. The other question is what you want to do with it, a healthy check would mean to check if everything is ready and loaded, maybe litellm provides a own health check (e.g., maybe one of them fits the behaviour: https://docs.litellm.ai/docs/proxy/health)

utils/rpcs/litellm/main.py
```diff
@@ -0,0 +1,84 @@
...
+
+            response = litellm.completion(
+                model=model,
+                messages=messages,
+                **params
+            )
+
+            result = {
+                "success": True,
+                "data": {
```
dennis-zyska: this I don't understand. It just copies the `response` into `data` (which may delete additional information) and adds a `success`, can't we just give back the whole response, log it in CARE and send relevant information to the frontend (may be only the `choices` here)

dennis-zyska (changes requested):

backend/webserver/rpcs/liteLLMRPC.js
akash9676: added

utils/rpcs/litellm/main.py
akash9676: I have to change this for all the other RPCs too right?? changing call to healthy

backend/webserver/services/nlp.js
akash9676: i have made the changes accordingly it i created another file in frontend so we can call the AI features without any UI required directly in any component

utils/rpcs/litellm/main.py
akash9676: yes we can do that

frontend/src/plugins/ai.js
```diff
@@ -0,0 +1,93 @@
...
+            clearTimeout(timer);
+
+            if (!response) {
+                reject(new Error("No response received from AIService"));
+                return;
+            }
+            if (response.success) {
+                resolve(response.data);
+            } else {
+                reject(new Error(response.message || "AIService request failed"));
```
dennis-zyska, line 90: should we send also on an abort message to the backend and forward to litellm, as the user doesn't expect a response here again and want to be able to rerequest it, the original request needs to be aborted

dennis-zyska (approved):

frontend/src/plugins/ai.js
akash9676: Yes but it is complicated to implement aborting and re requesting LLMs. i think we need to create our own req id tracking across end to end and cleanups etc...

there is also Litellm builtin cancel but that does not work with our socket io.

dennis-zyska: We can implement it until the RPC, so sending from frontend the timeout information to the backend (here we will log it), then sending at the same time the request back to the RPC that calls abort() on the python code. This code returns back "Not implementend" as an error which is also logged in the database. Then we can think about it later, be prepared it appropriately

PR #165, opened 2026-04-17
Feat 126 sidebar redesign

### LINKED ISSUE #126: [RESEARCH] Sidebar navigation restructuring
## Context

With the upcoming **Assignment feature**, another dashboard/page will likely be added to the application.  
The current sidebar navigation already contains many entries and may become harder to navigate as more sections are introduced.

To keep the interface clean and scalable, it might be useful to reconsider how navigation items are structured in the sidebar.

This issue collects a few **possible navigation patterns** that could help organize the sidebar without overloading it.

### Option A — Collapsible groups

Navigation items are grouped into expandable sections.

Example:

Image

Pros
- Clear structure
- User friendly

Cons
- Sidebar can still become long if many sections are expanded

### Option B — Sidebar with secondary flyout menu
Top-level items open a secondary navigation panel containing related pages.

Example:

Image

Pros
- Keeps the sidebar visually shorter
- User friendy

Cons
- Hover-based interaction may not work well on touch devices (click would be more reliable)
- Requires an additional sidebar panel, which may make the interface feel visually cluttered

### Option С — Icon-based primary sidebar with secondary navigation
Example:

Image

Pros:
- Keeps the main sidebar very compact

Cons:
- Icons may not always be self-explanatory
- Users need to interact with the sidebar to discover available pages

### Option D — Drill-down (folder-style) navigation

Sections behave similar to folders in a file system.  
Clicking a section opens its contents in the same navigation area.

The current level is indicated by path (e.g. Home > Dashboard > Studies) and a back button allows returning to the previous level.

Example:

Image

Pros
- Very clean sidebar (no long navigation list)
- Familiar mental model similar to file explorers
- Good for organizing many subsections
- Clear hierarchy through path

Cons
- Requires additional navigation steps (drilling down and going back)
- Users cannot see sibling sections at the same time
- May feel slower for frequent navigation between pages

## TODO
- Identify possible sidebar groups
- Define group names
- Decide how existing sections should be distributed between them

NAO-200:
Closes #126

## Changes
- dashboards are grouped into custom visual subgroups defined in the component
- on dashboard change, only the subgroup of the current route stays open
- collapsed mode now shows subgroup icons instead of only hiding text
- collapsed/closed subgroups support hover preview navigation
- active subgroup state is reflected visually in both expanded and collapsed modes
- subgroup arrows are animated and their position is stabilized
- sidebar scrolling and collapsed behavior were improved


frontend/src/components/dashboard/navigation/Sidebar.vue
```diff
@@ -58,112 +161,487 @@
...
+
   data() {
     return {
       version: APP_VERSION,
-    }
+      isCollapsed: false,
+
+      // Maps dashboards to visual sidebar subgroups
+      // Only elements listed here will be rendered inside the grouped sidebar
+      sidebarSubgroupConfig: {
```
dennis-zyska: This should be in the database by using the `nav_group` table: 

Currently we only have this in the table: 
```
const basicGroups = [
    {'name': "Default",},
    {'name': "Admin", "admin": true}
]
```
https://github.com/UKPLab/CARE/blob/dev/backend/db/migrations/20221010104856-basic-nav.js

We can reuse it, but need to adapt a bit the frontend, that these groups are used here instead where they are used now.

dennis-zyska (changes requested):

NAO-200:
As discussed in the last meeting, the group should be highlighted when its preview is open. Instead of relying on the arrow animation, I added a white highlight to the group. This also works in collapsed mode, where the arrow is not visible

backend/db/migrations/20260425001237-restructure-nav_group.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+const NEW_GROUPS = [
+  { name: 'Home', icon: 'house', order: 1 },
+  { name: 'Study', icon: 'book', order: 2 },
+  { name: 'Manage', icon: 'briefcase', order: 3 },
+  { name: 'Settings', icon: 'sliders2', order: 4 },
+  { name: 'AI', icon: 'code-slash', order: 5 },
```
dennis-zyska: we decided to use the robot icon right?

dennis-zyska (changes requested):

dennis-zyska (approved):

PR #181, opened 2026-04-26
Feat 175 export import generalization

karimouf:
## Export/Import Generalization

Refactored the export and import workflow into fully generic, reusable modal components (`ImportFormatModal`, `ExportFormatModal`) and integrated them across all major template (not user data) dashboard entities, replacing per-component implementations.

---

### New User Features

- **Export** any workflow, tag set, configuration, study template, or template directly from the dashboard table row via a download button
- **Import** any of the above entities from the dashboard header via an upload button
- **Format choice**: both export and import supports a number of formats import: (json, yml) export:(json, yml, text, csv)
- **Import wizard**: multi-step flow — file selection → data preview with column mapping → item selection → per-item success/failure feedback before committing

---

### New Dev Features

- `ImportFormatModal.open(table, childTable, { overrides, socket })` — generic API for any table:
  - `table`: table model for the imported data
  -  `childTable`: child table(if exists) for child table model like study_steps to studies for example
  - `overrides`: force-merge fields into every imported record (e.g. `{ template: true }`)
  - `socket.name`: emit to a custom socket event instead of `appDataUpdate`
  - `socket.dataKey`: nest item payload under a specific key, or spread flat if not used
  - `socket.extra`: inject extra top-level fields into the socket payload
- `ExportFormatModal.open(id, table, childTable, tableOptions, childTableOptions)` — generic export API:
  - **`id`** — primary key of the record to export; pass `null` to export all records currently loaded in the table
  - **`table`** — Vuex table name to read parent records from (e.g. `"workflow"`, `"tag_set"`); maps to `table/{table}/getFiltered` in the store
  - **`childTable`** — Vuex table name for records to nest inside each parent (e.g. `"workflow_step"`, `"tag"`); children are matched via a foreign key auto-derived from the parent name (e.g. `"tag_set"` → `tagSetId`); pass `null` when no nesting is needed
  - **`tableOptions`** — reserved for future parent-table configuration; pass `null` for now
  - **`childTableOptions.key`** — overrides the property name used to nest the children array in the exported object (e.g. `"tags"` instead of `"tag"`); defaults to the raw `childTable` name when omitted

---

### Improvements

- `UploadModal` is now documents-only (PDF) — configuration upload logic was removed and replaced by `ImportFormatModal`
- Consistent import/export UX and behavior across all dashboard views (Workflows, Tags, Configurations, Study Templates, Templates)

---

### Known Limitations

- Export reads from the client-side Vuex store — records not yet loaded into the store (e.g. due to pagination or lazy subscription) will not appear in the export file


frontend/src/assets/utils.js
```diff
@@ -94,34 +94,51 @@ export function objectsToTXT(objs) {
     ).join("\n\n");
 }
 
+/**
+ * Maps each supported export format to its serializer and MIME type.
+ * This is the single source of truth for downloadObjectsAs and getSupportedExportFormats.
+ */
+const EXPORT_FORMATS = {
+    csv:  { serialize: objectsToCSV,  mimeType: "text/csv",            label: "CSV Format",  icon: "filetype-csv",  description: "Comma-separated values, compatible with spreadsheets", extensions: [".csv"] },
```
dennis-zyska: can we format that a bit better, such that it is more easy to read?

frontend/src/basic/modal/ExportFormatModal.vue
```diff
@@ -0,0 +1,164 @@
...
+    return {
+      filterId: null,
+      table: null,
+      childTable: null,
+      childTableOptions: null,
+      tableOptions: null,
+    };
+  },
+  methods: {
+    /**
```
dennis-zyska, line 73: This docstring looks really ugly

frontend/src/basic/modal/ImportFormatModal.vue
```diff
@@ -0,0 +1,397 @@
...
+    acceptString() {
+      return this.acceptedExtensions.join(",");
+    },
+    steps() {
+      const itemLabel = this.table
+        ? this.table.replace(/_/g, " ")
+        : "Item";
+      return [
+        { title: "File Selection" },
+        { title: `${itemLabel} Selection` },
```
dennis-zyska, line 125: The itemLabel in the study import is `study`, which is why in the stepper in the frontend it looks like `study Selection`, but it would be better "Study Selection"

frontend/src/basic/modal/ImportFormatModal.vue
```diff
@@ -0,0 +1,397 @@
...
+        hover: true,
+        selectableRows: true,
+        scrollY: true,
+        search: true,
+      };
+    },
+  },
+  methods: {
+    /**
+     * Opens the import modal.
```
dennis-zyska, line 185: here the docstring as well, do we need all the examples etc.?

frontend/src/components/dashboard/documents/UploadModal.vue
dennis-zyska, line 1: What are the reasons for those removements?

frontend/src/components/dashboard/study/SavedTemplatesModal.vue
```diff
@@ -26,6 +26,20 @@
         name="Create Template"
         @click="createTemplate"
       />
+      <BasicButton
```
dennis-zyska: The buttons locations needs to improved, any idea?

dennis-zyska (changes requested):
- I tried to generate a study template and to download everything, but I get a TypeError in the browser console ("Cannot read properties of null (reading 'key')")
- Please also have a look into the lint problems, you can see it here in GitHub in the "Changes" tab of the PR

dennis-zyska (approved):

PR #184, opened 2026-04-28
Feat 160 create document metadata

pdji1602003:
## Main Description

This pull request contains the metadata-importing feature as shown in the following screen shot.

sc

New User Features

1. User can navigate to Assignment dashboard > Manage > Upload Metadata (Please make sure to create assignments before using this feature)
2. In the Import Metadata Modal, you can upload either a csv file or a json file. 
3. At the Target step, for now, CARE only offers Assignment for Target type.
4. Choose an assignment that your want to import the metadata to.
5. At Mapping step, for Primary Key Mapping, choose Username for Source key and extId for Match against.
6. At Metadata Mappings, user can choose what metadata they want to save into the database and under what name (Target metaKey).
7. At Preview and then Result, user can preview what information will be saved and the result returned from the db.


backend/webserver/sockets/document.js
```diff
@@ -839,6 +839,471 @@ class DocumentSocket extends Socket {
...
+    }
+
+    /**
+     * Import metadata rows for existing submissions in one assignment.
+     *
+     * @param {Object} data
+     * @param {Object} options
+     * @returns {Promise<Object>}
+     */
+    async importMetadata(data, options = {}) {
```
theultimatewasteofspace: This is callable by any connected user, but I don't see an authorization check on importMetadata or previewMetadataImport. Should this be admin-only or restricted to the assignment owner?

backend/webserver/sockets/document.js
```diff
@@ -839,6 +839,471 @@ class DocumentSocket extends Socket {
...
+        const submissionByEmail = new Map();
+
+        for (const submission of submissions) {
+            const owner = usersById.get(Number(submission.userId));
+            if (!owner) {
+                continue;
+            }
+
+            if (owner.extId != null) {
+                const extId = Number(owner.extId);
```
theultimatewasteofspace: I am not sure if extIds are guaranteed to be numeric across all auth sources but Number(owner.extId) means a non-numeric extId becomes NaN, which might not match in the Map. If they're always numeric, please ignore this.

theultimatewasteofspace: Never mind, its good, I checked.

backend/webserver/sockets/document.js
```diff
@@ -839,6 +839,471 @@ class DocumentSocket extends Socket {
...
+    async attachMappedMetadataToDocuments(data, options = {}) {
+        let writes = 0;
+
+        for (const documentId of data.documentIds) {
+            for (const mapping of data.mappings) {
+                if (!Object.prototype.hasOwnProperty.call(data.row, mapping.sourceField)) {
+                    continue;
+                }
+
+                await this.models["document_metadata"].upsertByDocumentAndKey({
```
theultimatewasteofspace: Since there's no unique index on (documentId, metaKey), could two imports at once insert duplicate rows for the same doc+key? Maybe a non-issue if imports are always sequential. Just checking.

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
```diff
@@ -0,0 +1,755 @@
...
+        return acc;
+      }, new Set())];
+      return { rows, fields };
+    },
+    parseCsv(file) {
+      return new Promise((resolve, reject) => {
+        Papa.parse(file, {
+          header: true,
+          skipEmptyLines: true,
+          complete: ({ data, meta }) => {
```
theultimatewasteofspace: parseCsv reads { data, meta } from Papa but not errors. Papa doesn't throw on a malformed CSV, so a broken file could silently drop rows without warning. Might be worth surfacing results.errors if it's non-empty?

backend/db/models/document_metadata.js
```diff
@@ -0,0 +1,93 @@
...
+
+        /**
+         * Create or update one metadata entry for a document and key.
+         *
+         * The topic import stores one logical record per document/key pair.
+         * If the key already exists for the document, the entry is updated in place
+         * instead of creating duplicate active rows.
+         *
+         * @param {Object} data
+         * @param {number} data.documentId｀
```
theultimatewasteofspace: tiny typo "`"

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
```diff
@@ -0,0 +1,755 @@
...
+                :disabled="metadataMappings.length === 1"
+                @click="removeMappingRow(index)"
+              >
+                Remove
+              </button>
+            </div>
+          </div>
+        </div>
+
+        <!-- TODO: Metadata presets, not sure what options are common here, to be discussed -->
```
theultimatewasteofspace: just flagging the TODO incase the topic has already been discussed

frontend/src/components/dashboard/Assignments.vue
```diff
@@ -330,6 +348,12 @@ export default {
         case "inspectSubmissions":
           this.$refs.assignmentSubmissionsModal.open(data.params.id);
           break;
+        case "uploadMetadata":
+          this.$refs.assignmentMetadataModal.open(data.params.id);
+          break;
+        case "togglePublic":
```
theultimatewasteofspace, line 355: togglePublic calls this.togglePublic(...) but I don't see a togglePublic method defined or a button dispatching this action. I might have missed it please ignore if so.

backend/db/models/document_metadata.js
pdji1602003: Resolved in commit 1a3773fb.

backend/webserver/sockets/document.js
pdji1602003: Added authorization check in commit d2a04ba8a8e208e34a7391c4a8b3ba3fdb0cb0a5.

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
pdji1602003: Nice point! Already addressed this in commit 5a9f7d74.

backend/webserver/sockets/document.js
pdji1602003: Resolved the issue in commit bb3c8470.

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
pdji1602003: Resolved the issue in commit 5a9f7d74.

frontend/src/components/dashboard/Assignments.vue
pdji1602003: I think we need to check with @karimouf about this. 😬

backend/db/migrations/20260629162000-extend-document-metadata_active_unique_index.js
dennis-zyska, line 1: What is the purpose of this function?

backend/webserver/sockets/document.js
```diff
@@ -839,6 +839,522 @@ class DocumentSocket extends Socket {
...
+
+        if (await this.hasAccess("frontend.dashboard.assignments.edit")) return;
+
+        throw new Error("You do not have permission to import metadata for this assignment.");
+    }
+
+    /**
+     * Persist mapped metadata for all documents of one matched submission row.
+     *
+     * @param {Object} data
```
dennis-zyska: there are a lot of docstrings where variables doesn't have a description. Also the MetaData function should go into a new class in sockets with the same name as the database, otherwise this file has to many lines

backend/webserver/sockets/document.js
```diff
@@ -839,6 +839,522 @@ class DocumentSocket extends Socket {
...
+        return entries.length;
+    }
+
+    /**
+     * Normalize metadata mappings into a canonical backend shape.
+     *
+     * @param {Object[]} mappings
+     * @returns {Object[]}
+     */
+    normalizeMetadataMappings(mappings = []) {
```
dennis-zyska: functions that are not directly called by the websocket request should go into another file e.g. in the utils folder

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
```diff
@@ -0,0 +1,756 @@
...
+    submit-text="Close"
+    @submit="$refs.metadataImportStepper.close()"
+    @step-change="handleStepChange"
+  >
+    <template #title>
+      <span>Import Metadata</span>
+    </template>
+
+    <template #step-1>
+      <div class="p-3">
```
dennis-zyska: we should split this file, each step should have another subcomponent, please ask @melolw, otherwise the file has too many lines

dennis-zyska (changes requested):

frontend/src/components/dashboard/assignments/AssignmentMetadataModal.vue
pdji1602003: Got it! Have split the file in commit 72991e78... 93b3e682

backend/db/migrations/20260629162000-extend-document-metadata_active_unique_index.js
pdji1602003: @theultimatewasteofspace mentioned the two import race condition in https://github.com/UKPLab/CARE/pull/184#discussion_r3416522149, so I added this migration to make sure that the combination of a document and a metaKey can't have two active rows in this table. First, I remove the duplicate rows if they already exist and then I add unique index so that there won't be duplicates in the future.

backend/webserver/sockets/document.js
pdji1602003: Resolved in commit dba9b888.

backend/webserver/sockets/document.js
pdji1602003: Resolved in commit c801cf32.

backend/db/migrations/20260629162000-extend-document-metadata_active_unique_index.js
pdji1602003: Or actually I don't have to consider the two import race condition since it's not likely to happen?

dennis-zyska: I feels like a complicated way, why not having a shared primary key for both?

backend/webserver/utils/helper/documentMetadata.js
dennis-zyska, line 1: please move to backend/webserver/utils/helper/documentMetaData.js

frontend/src/components/dashboard/assignments/metadata/StepMapping.vue
```diff
@@ -0,0 +1,247 @@
...
+            :value="mapping.metaKey"
+            class="form-control"
+            list="metadata-key-presets"
+            placeholder="topic"
+            type="text"
+            @input="updateMappingField(index, 'metaKey', $event.target.value)"
+          >
+        </div>
+        <div class="mapping-col-action">
+          <button
```
dennis-zyska: Use Basic Button ;-)

dennis-zyska (changes requested):

backend/db/migrations/20260629162000-extend-document-metadata_active_unique_index.js
pdji1602003: Do you mean using (documentId, metaKey) as a composite primary key?
I checked the current implementation. There is no frontend action to delete document metadata (document_metadata would be overwritten if there is already one in the db), and I could not find any backend path that soft-deletes document_metadata rows either.
So if we do not need soft deletion for this table, a normal unique constraint on (documentId, metaKey) would be simpler than  a composite primary key.

pdji1602003: Resolved in commit 0606861b.

backend/webserver/utils/helper/documentMetadata.js
pdji1602003: Resolved in commit 324419fd.

frontend/src/components/dashboard/assignments/metadata/StepMapping.vue
pdji1602003: Resolved in commit 14363dc9

backend/webserver/sockets/document_metadata.js
```diff
@@ -0,0 +1,412 @@
...
+            mapping.metaKey,
+            `${mapping.metaKey}.sourceFile`,
+            `${mapping.metaKey}.sourceField`,
+        ]);
+
+        const existingMetadataRows = documents.length > 0
+            ? await this.models["document_metadata"].findAll({
+                where: {
+                    documentId: documents.map((document) => document.id),
+                    metaKey: metadataKeys,
```
dennis-zyska, line 227: This doesn't filter deleted:false, so rows are counted in existingByDocumentId even when deleted.

dennis-zyska (approved):
I will merge it now, maybe we can do a quick hotfix for my comment

backend/webserver/sockets/document_metadata.js
pdji1602003: I have done the hotfix and made a PR for this.

PR #189, opened 2026-04-28
Feat 54 adding active users monitoring

MRawhani:
## Description

We added real-time monitoring stats to the admin dashboard.

We added information about active sessions and active unique users.
We added metadata for each session, like socket ID, browser info, and connection start time.

We added a separate room for monitors - every admin who enters the dashboard joins it.

**New User Features**

- New modal for active sessions - opens when clicking the table button for all users,
  and also via a new icon button that shows a filtered active session table per user.
- New column showing whether each user row is active or inactive - sortable and filterable.

**New Dev Features**

- Added a `buildStats()` function to the `Socket` base class.
- Added handshake time registration in `Server.js` at connection time.
- Added subscribing/unsubscribing on mount/unmount and users table and handle them in `user.js` in backend.

**Improvements**

**Bug Fixes**

**Known Limitations**
Filter in table is not working

**Future Steps**
Add more monitoring stats if needed and display related monitors to other role of users.


backend/webserver/sockets/user.js
```diff
@@ -398,7 +409,53 @@ class UserSocket extends Socket {
...
+     * @socketEvent userMonitorUnsubscribe
+     */
+    async unsubscribeFromUserMonitor(data, options) {
+        if (!(await this.isAdmin())) throw new Error("Admin access required");
+        this.socket.leave(MONITOR_USERS_ROOM);
+    }
+
+
     init() {
+        this.browser = parseUserAgent(this.socket.request.headers["user-agent"]);
```
dennis-zyska: lets move this to the Socket Class to keep it here clean

backend/webserver/sockets/user.js
```diff
@@ -398,7 +409,53 @@ class UserSocket extends Socket {
...
+
+        // Broadcast on new connection; fires before Server.js removes the entry on disconnect
+        this.broadcastStats();
```
dennis-zyska, line 456: this I don't understand, so it triggers here the broadcast for others right? so why not doing it directly in the Server.js where the connection is established?

frontend/src/components/dashboard/Users.vue
```diff
@@ -260,8 +316,24 @@ export default {
       ];
     },
   },
+  sockets: {
+    monitorStatsUpdate(data) {
```
dennis-zyska: We should move this to the Vuex Store directly (like here https://github.com/UKPLab/CARE/blob/dev/frontend/src/store/modules/admin.js#L94), that reduces the risk of lost messages

frontend/src/components/dashboard/Users.vue
```diff
@@ -379,6 +454,16 @@ export default {
       const filename = new Date().toISOString().replace(/[-:T]/g, '').slice(0, 14) + '_users';
       downloadObjectsAs(this.usersExport, filename, "csv");
     },
+    openSessionDetailsModal(params = null) {
+      const userId = params ? params.id : null;
+      this.modals.activeSessions = true;
+      this.$nextTick(() => this.$refs.activeSessionsModal?.open(userId));
+    },
+    handleMonitorUpdate(data) {
```
dennis-zyska: and then just use computed directly from the vuex store here, instead of setting it here

dennis-zyska (changes requested):
Thanks, looks good. I made a few comments and one general comment. When I click twice on the green indicator in the first column of the user list, I got an error in the browser console: "Cannot read properties of null (reading 'closest')"

backend/webserver/sockets/user.js
MRawhani: committed in 2c3f79415db5e86810dd14d593f527c3ada5f9eb and f98222c87c1f0f60a2cce8c555125ac2eda8050b (fix import)

backend/webserver/sockets/user.js
MRawhani: It actually created inconsistencies in timing with the object.entries snippet. sometimes live update doesnt work if it is in server connection handler. 

I tried to await all the array, the problem became worse. and I avoided re coding the snippet not to cause any backward dependency problem in other places.

so for now i left it as it is.

frontend/src/components/dashboard/Users.vue
MRawhani: commited in f98222c87c1f0f60a2cce8c555125ac2eda8050b

frontend/src/components/dashboard/Users.vue
MRawhani: commited in f98222c87c1f0f60a2cce8c555125ac2eda8050b

MRawhani: for the "Cannot read properties of null (reading 'closest, I think it is related to the Tooltip component for any column with type icon. It is just we dont see it in other icons because their first click is already opening a modal

dennis-zyska (approved):

PR #194, opened 2026-05-05
Encryption of sensitive data

### LINKED ISSUE #156: [FEATURE] Encrypting sensitive data in the database

Currently, we only encrypt passwords in the database, which are stored as passwordHash. But there are other equally sensitive information like names, email ids, initial passwords (if it exists), etc. which need to be encrypted.

One way is to encrypt these data in the database itself. For that we would use pgcrypto, an SQL module. We would create a function specifically to encrypt data. then we would search for that data using lookup hashes for that. We do not need to create specific columns for those. the function will handle that in the database itself

Another way is using sequalize-encrypted, but it not advantageous as it would burden the backend more. It is a node.js package which does the same work as pgcrypto.

junaidferoz:
## Description
This PR introduces generic application-layer field-level encryption for sensitive user data using AES-256-GCM. Models declare `encryptedFields` and encryption/decryption is handled automatically on all reads and writes via hooks on `GlobalChangeTrackingPlugin`. Key rotation and automatic encryption state sync on startup are also included.

---

## New User Features
- Sensitive user profile and authentication values are now encrypted at rest in the database
- Encryption can be toggled via `ENCRYPTION_ENABLED` in `.env` — the server applies the change automatically on next server start

---

## New Dev Features

**`backend/utils/encryption.js`**
- `encrypt(plaintext, key?)` — AES-256-GCM encryption with Base64 serialisation; accepts an optional explicit key
- `decrypt(value)` — always uses the stored key

**Model-level**
- `encryptedFields` option on model definitions for declarative field-level encryption — any model that declares it is automatically covered by all bulk operations

**Bulk operations**
- `encryptAllModels(db)` / `decryptAllModels(db)` — bulk encrypt/decrypt all fields across all models that declare `encryptedFields`

**Key rotation**
- `reEncryptAllModels(db, newKey)` / `reEncryptValue(value, newKey)` — old key read from `encryption.key`, new key passed explicitly
- `backend/scripts/changeEncryptionKey.js` + `make change_encryption_key NEW_KEY=<hex>` — CLI for key rotation without server downtime

**Startup sync**
- `syncEncryptionState(db)` — compares `ENCRYPTION_ENABLED` against `encryption.state` on startup and triggers encrypt/decrypt automatically if the flag changed

**Migration**
- `20260612100001-encrypt-user-fields` simplified to use `encryptAllModels` / `decryptAllModels` — new models with `encryptedFields` are automatically covered without migration changes

---

## Improvements

The following `user` model fields are now encrypted at rest:

| Category | Fields |
|---|---|
| Profile | `firstName`, `lastName`, `email`, `initialPassword`, `salt` |
| Auth | `twoFactorOtp`, `totpSecret`, `samlNameId` |
| External identity | `orcidId`, `ldapUsername`, `extId` |

---

## Known Limitations
- `ENCRYPTION_ENABLED` is detected at server startup only — changing the flag requires a restart

---

## Future Steps
- Extend `encryptedFields` to other tables as needed — no migration or script changes required
- Add an admin UI toggle for `ENCRYPTION_ENABLED` that triggers the sync without a restart


backend/db/index.js
akash9676, line 3: Add your name in the author

backend/db/index.js
```diff
@@ -19,6 +19,17 @@ const config = {
   ...loadedConfig,
   hooks: {
     ...(loadedConfig.hooks || {}), // Preserve existing hooks if we should ever add any in the config
+    afterConnect: async (connection) => {
+      if (loadedConfig.hooks && typeof loadedConfig.hooks.afterConnect === "function") {
+        await loadedConfig.hooks.afterConnect(connection);
+      }
+      const encryptionKey = process.env.DB_ENCRYPTION_KEY;
+      if (!encryptionKey) {
+        return;
```
akash9676: add error msg the copilot suggestion.

backend/db/migrations/20260505121500-transform-user-encryption-db-only.js
```diff
@@ -0,0 +1,317 @@
...
+                    "lastPasswordResetEmailSent",
+                    "lastVerificationEmailSent",
+                    "twoFactorOtp",
+                    "twoFactorOtpExpiresAt",
+                    "twoFactorMethods",
+                    "totpSecret",
+                    "orcidId",
+                    "ldapUsername",
+                    "samlNameId"
+                FROM public.user_secure;
```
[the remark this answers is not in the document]
akash9676: i think it is a good recommendation.

backend/db/migrations/20260505121500-transform-user-encryption-db-only.js
```diff
@@ -0,0 +1,317 @@
...
+                    "deletedAt",
+                    "acceptedAt",
+                    "acceptDataSharing",
+                    "rolesUpdatedAt",
+                    "extId",
+                    public.decrypt_text("initialPassword") AS "initialPassword",
+                    "emailVerified",
+                    "emailVerificationToken",
+                    "resetToken",
+                    "lastPasswordResetEmailSent",
```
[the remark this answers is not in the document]
akash9676: can you ask dennis do we need to encrypt them as well and also for 2fa otp and totpsectret.

akash9676 (changes requested):

backend/db/migrations/20260505121500-transform-user-encryption-db-only.js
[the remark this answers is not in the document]
junaidferoz: okay. @dennis-zyska -> Do we need to encrypt these columns as well, as well as the 2fa and totp ones also?

backend/db/migrations/20260505121500-transform-user-encryption-db-only.js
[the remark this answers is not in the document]
junaidferoz: Okay, thank you for the idea

backend/db/index.js
junaidferoz: Sure.

junaidferoz: i have did this in the latest commit

backend/db/index.js
junaidferoz: Added this in the latest commit

dennis-zyska (approved):
@bingobongomann the code looks good, could you test it locally, if it works, we can merge

bingobongomann:
> @bingobongomann the code looks good, could you test it locally, if it works, we can merge

grafik

Looks good. The User Table of a newly set up server is encrypted. New encryption key was generated during the migration. 
Frontend works as normal.

backend/scripts/changeEncryptionKey.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+/**
+ * Re-encrypt all model tables that declare encryptedFields, using the new key.
+ * The current key is read from backend/encryption.key.
+ *
+ * Usage (via Makefile):
+ *   make rotate-key NEW_KEY=<64-char hex>
```
bingobongomann: Can we make it an option to generate a new key if none is provided? Instead of throwing an error if no key is provided, trigger a user input asking if a new key should be generated and only abort if "No" is chosen. <
Also you changed the name of the make command to "change_encryption_key" instead of "rotate-key"

bingobongomann (changes requested):
Implementation looks good, encryption and decryption work well and are correctly updating the DB when the env is changed. 
Left one comment about make change_encryption_key to add optional generation of a new key if none is provided

Makefile
bingobongomann, line 1: the discussed changes are all working as expected, well done. 
For the case that we want to backup the db while the instance is running we should add a make backup_db_decrypted that creates a decrypted db_dump without having to change the .env and restarting the backend

bingobongomann (changes requested):
Looks Good now! Encryption can be updated via command line. 

The only thing missing is the decrypted database backup command and then we can merge this

backend/utils/encryption.js
```diff
@@ -313,4 +324,67 @@ function generateEncryptionKey() {
...
+ * @param {object} db  The db object exported from backend/db
+ */
+async function syncHashColumns(db) {
+    if (process.env.ENCRYPTION_ENABLED !== 'true') return;
+    const { DataTypes } = require('sequelize');
+    const { sequelize, models } = db;
+    const qi = sequelize.getQueryInterface();
+
+    for (const [, Model] of Object.entries(models)) {
+        const uniqueFields = (Model.options?.encryptedFields || [])
```
bingobongomann: I think its working fine the way it is, but we are checking if Model.options is there and if its undefined/null we use an empy array [] and then run the filter and map on the empty array. 
Not sure how big the overhead of running filter on an empty array is ( probably not that much) but logically it is redundant

backend/utils/helper/encryption.js
```diff
@@ -0,0 +1,390 @@
...
+            );
+            for (const row of rows) {
+                const plaintext = decrypt(row[fieldName]);
+                await sequelize.query(
+                    `UPDATE "${tableName}" SET "${hashField}" = :hash WHERE "${pk}" = :id`,
+                    { replacements: { hash: hashForUnique(plaintext), id: row[pk] } }
+                );
+            }
+
+            if (isNew) {
```
bingobongomann, line 377: how are we handling the already existing unique columns? we add new unique constraints for the new hashFields, but what about if it is an already exisitng field that has a unique constraint. Like with the emails, do we remove the uniqueness constraint from that field? With what i understand about the encryption it should be possible to have overlap due to the random part  that gets added to the message before encrypting, so we cannot just leave it as is right?

.env
```diff
@@ -65,5 +65,7 @@ PG_STATS_MIN_AGE_MS=1000
 #   to parameterize the LIMIT.
 PG_STATS_TOP_N=10
 
+#enable or disable encryption
```
bingobongomann: We need to add a warning that changing this will cause the whole db to change once the backend is restarted

bingobongomann (changes requested):
sidecar backup looks good, adding constraints and hooks for BlindIndexes looks good too. 
Just need Documentation and some clarification about existing unique fields.

karimouf:
this is ready @bingobongomann

backend/db/migrations/20260612100001-encrypt-user-fields.js
```diff
@@ -0,0 +1,167 @@
...
+ * Also populates emailHash from the newly encrypted email value.
+ *
+ * Requires DB_ENCRYPTION_KEY to be set in the environment.
+ * Skips rows where the field already appears encrypted (safe to re-run).
+ */
+
+const { encrypt, getKey, initializeEncryptionKey, decrypt } = require('../../utils/helper/encryption');
+
+module.exports = {
+    async up(queryInterface) {
```
dennis-zyska, line 14: The `up()` and `down()` here have a lot of duplicate code. Can we reuse the code from `backend/utils/helper/encryption.js` instead? It already has `encryptAllModels`/`decryptAllModels` (via `_applyToAllModels`)

backend/db/models/user.js
```diff
@@ -602,6 +602,8 @@ module.exports = (sequelize, DataTypes) => {
             sequelize,
             modelName: "user",
             tableName: "user",
+            //Keys that require encryption unique set to false by default
+            encryptedFields: ['firstName', 'lastName', { name: 'email', unique: true }, 'salt', 'initialPassword', 'twoFactorOtp', 'totpSecret', 'orcidId', 'ldapUsername', 'samlNameId'],
```
dennis-zyska, line 606: I think we also need to encrypt `extId`, but it is currently an Integer. Maybe we should change it to String, but we need to make sure that it will still work everywhere, maybe need a parse to int when requesting Moodle (needs testing).

backend/db/plugins.js
```diff
@@ -1,3 +1,104 @@
...
+
+    // Encrypt changed fields on UPDATE
+    addHook(options.hooks, 'beforeUpdate', (instance) => {
+        for (const name of fieldNames) {
+            if (instance.changed(name)) encryptField(instance, name);
+        }
+    });
+
+    // Encrypt on bulk INSERT
+    addHook(options.hooks, 'beforeBulkCreate', (opts) => {
```
dennis-zyska, line 81: `addEncryptionHooks` is not registering `beforeBulkUpdate`, which is called by `Model.update(...)`, e.g. from `User.resetUserPwd()` (backend/db/models/user.js:511). It would write the `salt` in plaintext to the database.

dennis-zyska (changes requested):

PR #200, opened 2026-05-06
Feat 83 recording (socket profiler recording & replay)

### LINKED ISSUE #83: [FEATURE] Recording & Replay Tool for CARE Platform

Currently, stress testing the CARE platform requires manually running external JS replay scripts, which is error-prone and not accessible to all team members. There is also no built-in way to verify that study sessions still work correctly after platform updates.

A built-in request recording and replay tool in the Admin section of CARE. Admins can record frontend-backend interactions (HTTP and WebSocket), save them as reusable traces, and replay them with configurable concurrency and speed, with a real-time monitoring dashboard and trace import/export for cross-instance use.

We considered keeping the external JS replay script and only adding a UI wrapper around it, but this limits flexibility to predefined operations and makes maintenance harder. Another option was a fully automated CI-based load testing pipeline, but this would not support ad-hoc recording of real user workflows.

## Additional context 

As described as below comment in details

theultimatewasteofspace:
## Description

Implements a Socket Profiler for CARE: a tool that records live Socket.IO
sessions, replays them against a running server, and uses those replays for
regression testing, load testing and analysis. Recording is started and stopped
by an admin from the Socket Profiler view; replay and analysis run from a CLI.

The motivation (Issue #83) is that CARE has no way to reproduce a real user
session against a changed backend. Manual clicking is currently the only
regression check for socket handlers, and it does not scale across the 46
documented user stories.

## New User Features

- **Socket Profiler admin view** — start and stop a recording of a live session,
  and browse past recordings in a table.
- **Recording detail modals** — inspect a recording's traces, per-event
  breakdown and run results without leaving the view.

## New Dev Features

- **Session recorder** — server-wide `onAny` capture of socket events into the
  `recording` table, with per-trace payload and timing.
- **Replayer** — replays a recorded session against a target server, remapping
  recorded hashes to live ones so create-then-open chains survive replay.
- **CLI with five modes**, run via `npm run perf` from `backend/`. Every mode
  exits with a meaningful status code, so all five are usable in CI:
  - `regression` — replay the story suite and report per-story pass/fail.
  - `inspect` — report a plain distribution of action types per recording
    (count and percentage), so a regression pass can be checked for what it
    actually exercised. Also flags recordings truncated by a page navigation.
    Informational; always exits 0.
  - `ramp` — escalate concurrency (N, 2N, 3N…) until a hard failure or the
    iteration cap, then report the degradation curve. Exits 0 if it reached the
    cap clean, 1 if it broke early.
  - `ceiling` — climb concurrency by `--step` with no cap until a stop condition
    trips (trace failure or p95 latency over threshold), and report the maximum
    concurrency the server sustained.
  - `soak` — hold a fixed concurrency for a duration, sampling each batch, and
    report whether the platform drifts over time. Exits 0 if stable, 1 if
    latency creep or failure accumulation is detected.
- **Supporting scripts** — `perf-metrics`, `report` and `trace-stats`, shared
  across modes for metric sampling, readable report output and trace-type
  breakdowns.
- **Regression suite** — 46 user stories under `backend/tests/regression_stories/`,
  each self-contained so it survives replay. 44 of 46 pass on a clean DB.
- **CLI flags** — `--fail-threshold` (0–100) and `--continue-on-failure`, with
  per-mode defaults; regression reports a per-story summary rather than aborting
  on the first failure.
- Replay target is configurable via `CONTENT_SERVER_PORT` rather than hardcoded.
- Story files are marked `linguist-generated` in `.gitattributes` to keep the
  diff reviewable.

## Known Limitations

- **The suite is single-use per database.** Soft deletes do not return sequence
  numbers, so a story cannot be replayed twice against the same DB. A clean
  database (`make clean && make init`) is required per full run.
- **Two stories are not yet replayable** and sit in `suite-needs-rework/`:
  `525-bulk-assignment` and `595-submit-multiple-times`.
- **Local auth is not captured.** Login, register, logout and 2FA go over
  Express rather than the socket, so they never appear in a recording.
- **Recording is server-wide.** If any participant's socket drops mid-recording,
  the whole recording stops and is flagged `disconnected`.
- **Traces that receive no acknowledgement count as failures.** Three guest
  annotator stories fail on the `collabUnsubscribe` trace, which has no backend
  handler, so the acknowledgement times out. All real mutations in those stories
  pass.
- **Regression is pass/fail only** — a run is not compared against any stored
  baseline.

## Future Steps

- Baseline comparison for regression runs, so a run can be judged as slower or
  more failing than a saved reference rather than only pass/fail.
- Decide how no-acknowledgement traces should be scored: pass, fail, or a
  distinct category. Passing timeouts wholesale would hide genuinely hanging
  handlers.
- Rework the two stories currently in `suite-needs-rework/`.
- Confirm the intended behaviour when a participant disconnects mid-recording:
  stop everything (current), keep partial traces, or stop only if the recording
  owner drops.


backend/db/migrations/20260412231119-create-trace.js
```diff
@@ -0,0 +1,86 @@
...
+      },
+      action: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      payload: {
+        type: Sequelize.JSONB,
+        allowNull: true,
+      },
+      direction: {
```
melolw, line 48: if i read this, i see direction is true or false, but don't know what true or false is meaning in this case. maybe add a comment that explains true is frontend to backend and false is backend to frontend (or other way around if you did it the other way) so that people know when they see it :)

backend/db/models/recording.js
```diff
@@ -0,0 +1,45 @@
...
+                foreignKey: "recordingId",
+                as: "traces",
+            });
+        }
+    }
+
+    Recording.init(
+        {
+            name: DataTypes.STRING,
+            status: DataTypes.STRING,
```
melolw: maybe you could make this an enum, with a fixed set of possible states? would make it a bit more robust imo

backend/db/models/recording.js
```diff
@@ -0,0 +1,45 @@
+"use strict";
```
melolw, line 1: for both recording.js and trace.js I don't see any validators, i feel like status should probably always be non-null as an example

backend/db/models/trace.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    Trace.init(
+        {
+            recordingId: DataTypes.INTEGER,
+            userId: DataTypes.INTEGER,
+            socketId: DataTypes.STRING,
+            action: DataTypes.STRING,
+            payload: DataTypes.JSONB,
+            direction: DataTypes.BOOLEAN,
```
melolw: here as well a comment that tell us what true and what false is

backend/db/models/recording.js
```diff
@@ -0,0 +1,45 @@
...
+const MetaModel = require("../MetaModel.js");
+
+module.exports = (sequelize, DataTypes) => {
+    class Recording extends MetaModel {
+        static autoTable = true;
+        static fields = [];
+        static publicTable = true;
```
melolw: you mentioned why this is needed in your description, you could add this explanation here as a comment as well

backend/webserver/replay/auth.js
```diff
@@ -0,0 +1,117 @@
...
+        const match = cookie.match(/connect\.sid=s%3A([^.]+)\./);
+        if (match) {
+            await server.db.sequelize.query(
+                `DELETE FROM "Sessions" WHERE "sid" = :sid`,
+                { replacements: { sid: match[1] } }
+            );
+        }
+        client.disconnect();
+    } catch (err) {
+        // best-effort cleanup
```
melolw: not sure what error could happen here, but if you have a try/catch, that means you think it could fail right? so a console.warn would be good to avoid silent fails

backend/webserver/sockets/recorder.js
```diff
@@ -0,0 +1,244 @@
...
+        for (const socketId of Object.keys(this.server.availSockets)) {
+            const recorder = this.server.availSockets[socketId]["RecorderSocket"];
+            const included = recorder ? recorder.isSessionIncluded(recorder.socket.id) : "no recorder";
+            if (recorder && recorder.isSessionIncluded(recorder.socket.id)) {
+                recorder.attachListeners();
+            }
+        }
+    }
+
+    async stopRecording(data, options) {
```
melolw, line 215: maybe this should check the authentification as well?

backend/webserver/sockets/replayer.js
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * @param {Array<number>} recordingIds - Recordings to pool
+     * @returns {Promise<{sessions: Array, userMap: Map}>} Pooled sessions and the user lookup needed for replay
+     */
+    async buildSessionPool(recordingIds) {
+        const sessions = [];
+        const userIdSet = new Set();
+
+        for (const recordingId of recordingIds) {
+            const recording = await this.models['recording'].findByPk(recordingId);
```
melolw: you could probably do one db query like: findAll({ where: { id: recordingIds } }) outside of the loop instead of doing for each recordingId another db query

frontend/src/components/dashboard/socketprofiler/RecordingModal.vue
```diff
@@ -0,0 +1,223 @@
...
+      // later sorts or filters the table.
+      const ordered = [...this.allTraces].sort((a, b) => {
+        const aMs = a.startTime ? new Date(a.startTime).getTime() : 0;
+        const bMs = b.startTime ? new Date(b.startTime).getTime() : 0;
+        return aMs - bMs;
+      });
+      return ordered.map((t, index) => ({
+        ...t,
+        sequence: index + 1,
+        directionLabel: t.direction ? '→ backend' : '← frontend',
```
melolw: it should be → backend and → frontend with the same direction arrow to represent correctly both direction. Currently both mean "to backend"

frontend/src/components/dashboard/socketprofiler/ReplayResultsModal.vue
```diff
@@ -0,0 +1,246 @@
...
+
+        <div v-for="(sessionResult, idx) in iteration.results" :key="idx" class="mb-3">
+          <p class="mb-1 fw-bold">
+            {{ sessionResult.userName }} (User ID: {{ sessionResult.userId }})
+            <span class="text-muted small">— from {{ sessionResult.recordingName || 'Recording ' + sessionResult.recordingId }}</span>
+            — {{ sessionResult.passed }}/{{ sessionResult.total }} passed
+          </p>
+          <BasicTable
+            :columns="traceColumns"
+            :data="mergeTraces(sessionResult, iteration.level, idx)"
```
melolw: shouldnt this be called in computed once and not in the template part? not sure tho

frontend/src/components/dashboard/socketprofiler/ReplayResultsModal.vue
```diff
@@ -0,0 +1,246 @@
...
+                class="btn-outline-secondary btn-sm"
+                text="Close"
+                icon="x"
+                @click="clearSelection"
+              />
+            </div>
+            <div v-if="selectedTrace.dbChanges && selectedTrace.dbChanges.length > 0">
+              <BasicTable
+                :columns="dbChangeColumns"
+                :data="selectedTrace.dbChanges.map((c, i) => ({
```
melolw: same here, shouldnt this be in computed so that its computed once?

melolw (commented):
great feature!
you should be careful when you add files, we apple users (🥲) have hidden files called .DS_Store that are added to the commit when adding every diffs without checking first. I saw 2 .DS_Store files in your pull request, not sure how Dennis handles it when merging but I simply try to never add them in my commits in the first place

backend/db/models/recording.js
theultimatewasteofspace: good idea! I went with an isIn validator + a constant instead of a real enum. It should have the same protection. I noticed we dont use enums anywhere in CARE and the status set's still growing (disconnected's coming), so 
i hesitated and went this way.

backend/db/models/recording.js
theultimatewasteofspace: added validators on both. status is non-null + isIn-checked, and required the fields that are always set at creation (recordingId, action, direction, timestamps). left payload, endTime, userId, socketId nullable on purpose cuz sometimes payload's legit empty for some events, endTime's null while recording, and userId/socketId aren't set on every path. good catch, Thanks! :)

backend/webserver/replay/auth.js
theultimatewasteofspace: good point, that was a silent fail waiting to happen. added a console.warn so if the session DELETE ever fails it'd leave orphan rows, so better to surface it. no logger in scope here since it's a plain function, hence console.warn

backend/webserver/sockets/recorder.js
theultimatewasteofspace: nice catch! it wasn't checking auth at all (neither was startRecording). added isAdmin() guards to both, following the same pattern as setting.js/user.js.

backend/webserver/sockets/replayer.js
theultimatewasteofspace: good idea! I left the per-recording trace query as-is since the grouping right after depends on it being per-recording.

backend/webserver/sockets/replayer.js
```diff
@@ -0,0 +1,457 @@
...
+        const tracesPerPass = pool.sessions.reduce((sum, s) => sum + s.traces.length, 0);
+        const totalTraces = tracesPerPass * (maxIterations * (maxIterations + 1) / 2);
+
+        const progress = this.makeProgressReporter(progressId, totalTraces);
+
+        for (let level = 1; level <= maxIterations; level++) {
+            // Iteration K runs K full copies of the pool. Total sockets = K * N.
+            // Sessions are picked with wraparound, so iteration K's list is
+            // pool[0], pool[1], ..., pool[N-1], pool[0], pool[1], ... (K times).
+            const totalSockets = level * N;
```
NAO-200, line 410: `MAX_CONCURRENCY` is enforced for runOneLevel(), but the scaling path appears to bypass that limit. Since `totalSockets` grows as `level * N`, a large enough pool can easily exceed the advertised maximum (e.g. 20 pooled sessions × 100 iterations = 2000 concurrent sockets while `MAX_CONCURRENCY` is 500).

Was this intentional? If `MAX_CONCURRENCY` is meant as a global safety limit , it may be worth validating

backend/webserver/sockets/replayer.js
```diff
@@ -0,0 +1,457 @@
...
+            // failures. Off by default (Infinity); the perf tool opts in.
+            const levelLatencies = [];
+            for (const r of levelResults) {
+                for (const l of (r.latencies || [])) levelLatencies.push(l.latency);
+            }
+            levelLatencies.sort((a, b) => a - b);
+            const p95 = levelLatencies.length
+                ? levelLatencies[Math.min(levelLatencies.length - 1, Math.floor(0.95 * levelLatencies.length))]
+                : 0;
+            const latencyExceeded = p95 > latencyThreshold;
```
NAO-200, line 432: `latencyThreshold` doesn't appear to be validated, unlike the other numeric inputs. If someone accidentally pass nan or another non numeric value, p95 > latencyThreshold will always be false, disabling the latency stop condition. Would it make sense to validate this parameter in replayRun() as maxIterations and singleLevel?

backend/webserver/replay/auth.js
```diff
@@ -0,0 +1,153 @@
...
+            },
+            type: server.db.sequelize.QueryTypes.INSERT,
+        }
+    );
+    // Replay clients mint their own session cookies, so this must be the same
+    // secret express-session verifies with in Server.js (#initSessionManagement)
+    // — otherwise the middleware rejects them and replay can't authenticate.
+    const secret = process.env.SESSION_SECRET;
+    if (!secret) {
+        throw new Error('SESSION_SECRET is not set — replay cannot mint session cookies');
```
NAO-200: The session has already been inserted into the table in lines above.
But if secret is missing, this throws before execution reaches the `try/catch` below with path that calls `deleteSession`. So newly created session row will remain in the table until it expires
Maybe it would make sense to validate secret before the insertion

backend/scripts/perf-auth.js
```diff
@@ -0,0 +1,59 @@
...
+    return match[0];
+}
+
+/**
+ * Verify a session cookie is a logged-in admin via /auth/check.
+ * @param {string} serverUrl
+ * @param {string} cookie - the "connect.sid=..." cookie
+ * @returns {Promise<Object>} the user object
+ */
+async function verifyAdmin(serverUrl, cookie) {
```
NAO-200: The name `verifyAdmin` suggests that this function verifies administrator privileges, but it currently only checks that the session is authenticated and returns data.user. if we only want to validate the session cookie, would a name like `verifyAuthenticatedSession` (or similar) better reflect its behavior? Or if this helper is expected to guarantee an admin session, it may be worth explicitly validating the returned user's role here.

backend/scripts/perf-ceiling.js
```diff
@@ -0,0 +1,188 @@
...
+            progressId,
+        }, 0);
+
+        ctx.socket.off('progressUpdate', onProgress);
+        process.stdout.write('\r' + ' '.repeat(50) + '\r');  // clear the progress line
+
+        if (!ack || !ack.success) {
+            await sampler.stop();
+            console.error('\nCEILING ERROR — replayRun failed: ' + (ack && ack.message));
+            capture.stop();d
```
NAO-200: `;d` looks like a typo

backend/scripts/perf-ceiling.js
```diff
@@ -0,0 +1,188 @@
...
+        level++;
+        const progressId = randomUUID();
+        const onProgress = (data) => {
+            if (data.id !== progressId) return;
+            const pct = data.total ? Math.floor((data.current / data.total) * 100) : 0;
+            process.stdout.write(`\r  level ${level} (concurrency ${concurrency}): ${pct}%   `);
+        };
+        ctx.socket.on('progressUpdate', onProgress);
+
+        const ack = await ctx.emitWithAck('replayRun', {
```
NAO-200: If `emitWithAck()` throws an error, we skup `ctx.socket.off('progressUpdate', onProgress)`. Maye we should wrap this section in try/finally to guarantee cleanup?

backend/scripts/perf/ceiling.js
```diff
@@ -0,0 +1,188 @@
...
+const { resolvePayload } = require('./perf-recordings');
+const { randomUUID } = require('crypto');
+const { MetricSampler } = require('./perf-metrics');
+const { printTraceStats } = require('./perf-trace-stats');
+const { saveResults, makeOutputCapture, saveReadableReport } = require('./perf-report');
+
+/**
+ * Ceiling finder: climb concurrency by --step each level, no cap, until a stop
+ * condition trips (trace failure or p95 latency over threshold). Reports the
+ * max concurrency the server sustained. Drives escalation one level at a time
```
NAO-200, line 12: The jsdoc doesn't seem to match the implementation. It says this function reports the maximum sustained concurrency, but it actually returns 0/1 as an exit code, while the measured value (`lastGood`) is only written to stdout and never returned. When someone would call this function, he would expect something like:
```js
const maxConcurrency = await runCeiling(...);
```
and receive the measured ceiling, but would instead get an exit code.

backend/scripts/perf/ceiling.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        const m = metrics(ack.data);
+        allLevels.push({ level, concurrency, ...m, results: ack.data.results });
+
+        // Pool-waiting: the DB pool backing up = pool exhaustion (the original
+        // failure). Read the peak since we started; if it climbed above 0, the
+        // pool saturated at this concurrency.
+        const peakWaiting = sampler.peakPoolWaiting();
+        const poolSaturated = peakWaiting > poolWaitingBaseline && peakWaiting > 0;
```
NAO-200, line 87: Unless I'm missing something, `poolWaitingBaseline` is initialized to 0 but never updated, so this condition is currently equivalent to peakWaiting > 0. Is the baseline intended to change between iterations, or can this comparison be simplified?

backend/scripts/perf-inspect.js
```diff
@@ -0,0 +1,72 @@
...
+    if (sessions.length) {
+        for (const session of sessions) {
+            analyzeRecording(session.recordingName || session.sessionKey, session.traces || []);
+        }
+        return 0;
+    }
+
+    // DB path: fetch each recording's traces from the server by id.
+    for (const id of recordingIds) {
+        const ack = await ctx.emitWithAck('recordingGetTraces', { id });
```
NAO-200: Similar to the `runCeiling()` path, this only handles logical failures (!ack / !ack.success). If it will throw an error (e.g. due to socket failure), the exception will abort the entire inspection loop and no remaining recordings will be processed

backend/scripts/perf-ramp.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+
+const { resolvePayload } = require('./perf-recordings');
+const { randomUUID } = require('crypto');
+const { printTraceStats, printCulprit } = require('./perf-trace-stats');
```
NAO-200: `printCulprit` appears to be unused in this file. If it's no longer needed, it can probably be removed

backend/scripts/perf-ramp.js
```diff
@@ -0,0 +1,158 @@
...
+        process.stdout.write(`\r  progress: ${pct}% (${data.current}/${data.total} traces)   `);
+    };
+    ctx.socket.on('progressUpdate', onProgress);
+
+    // Sample server vitals (memory + DB pool) during the whole ramp climb.
+    const sampler = new MetricSampler(ctx.emitWithAck, 1000);
+    sampler.start();
+
+    // The previously broken/duplicate emitWithAck call, now restored
+    const ack = await ctx.emitWithAck('replayRun', {
```
NAO-200: same as in `runCeiling()`

backend/scripts/perf/regression.js
```diff
@@ -0,0 +1,107 @@
...
+        sessions,
+        timingMode: 'fast',
+        continueOnFailure: true,   // run all stories so we report every failure, not just the first
+        maxIterations: 1,          // once — no scaling
+        ackTimeout: cfg.ackTimeout,
+    },0);
+
+    if (!ack || !ack.success) {
+        console.error('REGRESSION ERROR — replayRun failed: ' + (ack && ack.message));
+        return 1;
```
NAO-200, line 59: capture.start() is called at the beginning of the function, but capture.stop() is not reached on this early return path. Is that intentional? 
Also `emitWithAck()` is only handled when it resolves successfully. If it rejects, the exception propagates immediately and capture.stop() is skipped as well

backend/scripts/perf-regression.js
```diff
@@ -0,0 +1,107 @@
...
+
+    // ack.data is the iteration results array. maxIterations=1 => one iteration.
+    const iterations = ack.data || [];
+
+    // Group results per recording (per user story) so we can report which
+    // stories pass and which fail — the "is this version stable?" verdict.
+    const stories = new Map();  // key: recording name/id -> { total, failed, errors[] }
+    for (const iter of iterations) {
+        for (const session of (iter.results || [])) {
+            const key = session.recordingName || ('recording ' + session.recordingId);
```
NAO-200: If `recordingName` is not guaranteed to be unique, different recordings could be merged together here

backend/scripts/perf/utils/report.js
```diff
@@ -0,0 +1,95 @@
...
+ * @returns {string|null} the .txt path, or null
+ */
+function saveReadableReport(jsonPath, text) {
+    try {
+        if (!jsonPath) return null;
+        const txtPath = jsonPath.replace(/\.json$/, '.txt');
+        fs.writeFileSync(txtPath, text, 'utf8');
+        return txtPath;
+    } catch (err) {
+        return null;
```
NAO-200, line 94: `saveReadableReport()` silently ignores write failures, unlike saveResults()

backend/scripts/perf/soak.js
```diff
@@ -0,0 +1,191 @@
...
+            progressId,
+        }, 0);
+
+        ctx.socket.off('progressUpdate', onProgress);
+        process.stdout.write('\r' + ' '.repeat(40) + '\r');
+
+        if (!ack || !ack.success) {
+            console.error('SOAK ERROR — replayRun failed: ' + (ack && ack.message));
+            await sampler.stop();
+            return 1;
```
NAO-200, line 67: `capture.stop()` is skipped on this early return as well

backend/scripts/perf-soak.js
```diff
@@ -0,0 +1,191 @@
...
+        sampleNum++;
+        const progressId = randomUUID();
+        const onProgress = (data) => {
+            if (data.id !== progressId) return;
+            const pct = data.total ? Math.floor((data.current / data.total) * 100) : 0;
+            process.stdout.write(`\r  sample ${sampleNum}: ${pct}%   `);
+        };
+        ctx.socket.on('progressUpdate', onProgress);
+
+        const ack = await ctx.emitWithAck('replayRun', {
```
NAO-200: If `emitWithAck()` rejects instead of returning an unsuccessful acknowledgement, `progressUpdate` is never unsubscribed and `sampler.stop()` is skipped as well

backend/scripts/perf-ceiling.js
```diff
@@ -0,0 +1,188 @@
...
+                const pg = sampler.pgStatsSummary();
+                if (pg) {
+                    console.log(`  Postgres: deadlocks +${pg.deadlocksDelta}, rollbacks +${pg.rollbacksDelta}, peak lock-waits ${pg.peakLockWaits}` +
+                        (pg.minCacheHit != null ? `, min cache-hit ${pg.minCacheHit.toFixed(1)}%` : ''));
+                }
+            }
+            printTraceStats(ack.data.results, { title: `Trace breakdown at level ${concurrency} (each action's OWN stats):` });
+
+            // Name the likely culprit, scoped to WHY it stopped, with clear p95 labeling.
+            const culpritStats = require('./perf-trace-stats').traceStats(ack.data.results);
```
NAO-200: The culprit selection performed largely duplicates the logic already encapsulated in printCulprit() in `perf-trace-stats.js`. Maybe we can reuse it?

backend/db/migrations/20260506220447-move-socket-profiler-to-manage.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+/**
+ * Move Socket Profiler nav element into the Manage group.
+ *
+ * The original `extend-nav-socket-profiler` migration (March 22, 2026)
```
NAO-200: Is there a reason this is implemented as a separate migration instead of updating the original `extend-nav-socket-profiler` migration? Usually, after a migration has been merged to main/dev branch, adding a new migration is the right approach. However, since this branch hasn't been merged yet, perhaps the original migration could simply be updated to keep the migration history simpler

frontend/src/components/dashboard/socketprofiler/StartRecordingModal.vue
```diff
@@ -0,0 +1,252 @@
...
+    startButtonText() {
+      return "Record " + this.selectedSessions.length + " Session(s)";
+    },
+    allExcludeEvents() {
+      return [...this.excludeEvents, ...this.customExcludes];
+    },
+  },
+  sockets: {
+    sessionsChanged() {
+      if (this.isOpen) this.refreshSessions(false);
```
NAO-200, line 161: `isOpen` stays true if the modal closed via "Х" button or esc, since it is only reset in `abort()` or `confirm()`. As a result, `sessionsChanged` continues to refresh the session list after the modal has been closed. It might be worth resetting `isOpen` on the modal's `@hide` event as well

NAO-200 (commented):
It was genuinely interesting to read through your task and see how you implemented everything. I left a few comments with possible improvements, but overall amazing work! Also, don't forget to add your author signature to the new files

backend/webserver/sockets/replayer.js
theultimatewasteofspace: Good catch. Root cause was in the CLI parser: parseInt(args['latency-threshold'], 10) || 1000 turned any non-numeric value (nan, etc.) into a silent 1000ms default, so a bad input neither errored nor disabled the stop and it quietly changed the threshold. Since latencyThreshold is only reachable via the CLI, fixed it at the arg boundary in perf-test.js alongside --max-iterations: bad values now error out before the run starts. Omitting the flag still gives the 1000ms default; passing a real number works as before.

backend/webserver/replay/auth.js
theultimatewasteofspace: Moved the SESSION_SECRET check above the INSERT so it fails before anything's written. The insert doesn't need the secret (only cookie signing does), so it's a clean reorder. good catch

backend/scripts/perf-auth.js
theultimatewasteofspace: You're right. VerifyAdmin only validated the session, didn't check the role. The actual admin gate is verifyAdminAccess(), which calls an admin-guarded socket handler before any replay work. Renamed this one to verifyAuthenticatedSession to match what it does, and added a guard for the user: null case (/auth/check returns 200 when logged out). Admin stays enforced at the socket.

backend/scripts/perf-ceiling.js
theultimatewasteofspace: Good catch. Listener leak on the error path. If emitWithAck threw, the .off('progressUpdate', ...) was skipped and the handler stayed attached. Wrapped the emit in try/finally so cleanup always runs. Same pattern was in perf-ramp.js and perf-soak.js, fixed those too.

backend/scripts/perf/ceiling.js
theultimatewasteofspace: Youre right. lastGood was only printed, never returned, and the @returns overclaimed. These mode fns return their value as the process exit code (perf-test.js does process.exit(await runCeiling(...))), so I couldn't just return the ceiling, that'd break the exit status. Changed it to return { code, maxConcurrency }: perf-test.js uses .code, and a programmatic caller now gets the measured ceiling. Doc updated to match.

backend/scripts/perf/ceiling.js
theultimatewasteofspace: Yes, baseline was declared but never assigned, so the check collapsed to peakWaiting > 0. It was meant to capture the pool's idle waiting before load, so saturation means "above resting" rather than "above zero" (some deployments idle non-zero). Wired it up to sample latestPoolWaiting() right after start. Unchanged on a pool that idles at 0.

backend/scripts/perf-inspect.js
theultimatewasteofspace: Good catch, a throw (vs a logical !ack.success) would abort the whole loop. Wrapped each fetch in try/catch that logs and continues, mirroring the existing !ack.success handling, so one unreadable recording doesn't kill the batch.

backend/scripts/perf-ramp.js
theultimatewasteofspace: Fixed. Wrapped the emit in try/finally so the listener and sampler are cleaned up on every exit path including a throw. Same fix applied to runCeiling() and runSoak().

backend/scripts/perf-ramp.js
theultimatewasteofspace: Right. printCulprit is used in the ceiling/inspect modes but not here. Removed it from this file's import; kept printTraceStats which is used.

backend/scripts/perf/regression.js
theultimatewasteofspace: Good catch on both. The bad-ack early return and a throw from emitWithAck both skipped capture.stop(), leaving the output tee attached. Added capture.stop() to the early return path and wrapped the emit in try/catch to stop it on a throw too. Didn't save a report on these paths since the replay never ran so there's no result to report, just the error, which is already logged. The normal pass/fail paths already stop+save correctly and are untouched.

backend/scripts/perf-regression.js
theultimatewasteofspace: Switched the key to a stable identity, recordingId for DB replay (groups a recording's sessions), falling back to sessionKey for file replay where there's no id. recordingName is now display-only.

backend/scripts/perf/utils/report.js
theultimatewasteofspace: Added a matching console.error before the return null so a failed readable-report write is now surfaced the same way, keeping the null return contract.

backend/scripts/perf/soak.js
theultimatewasteofspace: Added it before return 1, matching the fix in the regression path. The sampler is already stopped by the outer finally, and the success path's capture stop is untouched.

backend/scripts/perf-soak.js
theultimatewasteofspace: Both handled. The progressUpdate listener is now removed in a per-level try/finally around the emit, and sampler.stop() moved into an outer try/finally around the loop, so a throw from emitWithAck cleans up both. Same fix applied to ramp and ceiling.

backend/scripts/perf-ceiling.js
theultimatewasteofspace: Good catch. Replaced it with a printCulprit(results, { failed, poolSaturated, overallP95 }) call and imported it at the top.

backend/db/migrations/20260506220447-move-socket-profiler-to-manage.js
theultimatewasteofspace: You're right, i missed it. Updated the original migration.

frontend/src/components/dashboard/socketprofiler/StartRecordingModal.vue
theultimatewasteofspace: Added an @hide handler (BasicModal emits hide on every close path) that resets isOpen. Didn't route it through abort() since that calls modal.close(), which X has already triggered so onHide just resets the flag.

backend/db/migrations/20260506220447-move-socket-profiler-to-manage.js
dennis-zyska: so we can delete this file?

theultimatewasteofspace: yes, deleted it.

backend/scripts/perf-auth.js
```diff
@@ -0,0 +1,63 @@
...
+ * Authenticate as a configured admin via the real /auth/login route and
+ * return the session cookie for socket connections. No session forging — this
+ * is a normal login, so it doesn't depend on the session secret.
+ *
+ * @param {string} serverUrl - e.g. http://localhost:3001
+ * @param {string} userName - admin username (default seeded admin is "admin")
+ * @param {string} password - admin password (from config/env, never hardcoded)
+ * @returns {Promise<string>} The "connect.sid=..." cookie string
+ */
+async function loginAsAdmin(serverUrl, userName, password) {
```
dennis-zyska: Why is this function named `loginAsAdmin`, it is general and would work with any user name right? So we can call also the function more general

backend/scripts/perf/utils/auth.js
dennis-zyska, line 1: we should move all files into a subdirectory `./backend/scripts/perf`

backend/scripts/perf/ceiling.js
```diff
@@ -0,0 +1,195 @@
...
+            return { code: 0, maxConcurrency: lastGood };
+        }
+        concurrency += step;
+    }
+    } finally {
+        if (!stopped) await sampler.stop();
+    }
+}
+
+function metrics(lvl) {
```
dennis-zyska, line 191: Please always provide a docstring for function explaining what it is doing (counts also for the other functions)

backend/scripts/perf/utils/report.js
dennis-zyska, line 1: I think the directory needs some cleaning in general. If we have a subfolder perf, we don't need `perf` in the name for all files. In addition, there are more general files, like this report or the auth file, that is used in several other files. We should have another subfolder for this like `./scripts/perf/utils/`

backend/scripts/perf/utils/report.js
```diff
@@ -0,0 +1,96 @@
...
+ * pg reading), complete per-trace stats, config, and verdict.
+ *
+ * @param {string} mode - which mode ran (soak, ceiling, ramp, regression)
+ * @param {Object} cfg - the run's config
+ * @param {Object} payload - mode-specific data: { results, samples, sampler, verdict, extra }
+ * @returns {string|null} the file path written, or null on failure
+ */
+function saveResults(mode, cfg, payload = {}) {
+    try {
+        const dir = path.join(process.cwd(), 'perf-results');
```
dennis-zyska: the backend creates automatically a `log` directory in the backend. The advantage is, that we already adapted the docker-compose for it, so we should also save those results in the log directory, just create a subdirectory inside of this log directory for the performance results

backend/scripts/perf/soak.js
```diff
@@ -0,0 +1,198 @@
...
+const { MetricSampler } = require('./perf-metrics');
+const { printTraceStats, printMemoryCorrelation } = require('./perf-trace-stats');
+const { saveResults, makeOutputCapture, saveReadableReport } = require('./perf-report');
+
+/**
+ * Soak mode: hold a FIXED concurrency continuously for a duration, sampling
+ * each batch, and report whether the platform drifts over time (latency creep
+ * or failure accumulation at constant load). Reuses replayRun's singleLevel
+ * mode to run one fixed-concurrency batch per sample.
+ * @param {Object} cfg
```
dennis-zyska: it is not enough to state only the variable name, it doesn't explain what the function expect as content, please update the docstrings also in the other files

backend/scripts/perf/soak.js
```diff
@@ -0,0 +1,198 @@
...
+        const txt = saveReadableReport(saved, text);
+        console.log(`\n  results saved: ${saved}`);
+        if (txt) console.log(`  readable report: ${txt}`);
+    }
+    return code;
+}
+
+/**
+ * Compare early vs late samples to detect drift over time.
+ * @returns {number} exit code
```
dennis-zyska: here are also all other parameters missing in the docstring

backend/utils/replay-auth.js
dennis-zyska, line 1: we should move those two files into the `.backend/utils` folder as helpers

backend/webserver/sockets/recorder.js
```diff
@@ -0,0 +1,437 @@
...
+     * @throws {Error} If the caller is not an admin or no id is given
+     */
+    async getTraces(data, options) {
+        if (!(await this.isAdmin())) {
+            throw new Error("Admin access required");
+        }
+        if (!data || !data.id) throw new Error("Recording ID required");
+        const traces = (await this.models["trace"].getAllByKey("recordingId", data.id, options))
+            .sort((a, b) => a.id - b.id);
+        return traces.map(t => ({
```
dennis-zyska: I wonder why we need to change the array here? Can't we send the complete entry as it comes from the database? Usually we just sending copies of the complete set, reducing managable code as much as possible (this questions counts also for the other returns)

backend/webserver/Server.js
```diff
@@ -377,11 +377,24 @@ module.exports = class Server {
...
+                    await this.availSockets[socket.id][socketName].init();
+                })
+            );
+
+            // All per-socket handlers are now initialized and listening, so it's
+            // safe for clients (including replay clients) to start emitting events.
+            socket.emit("ready");
+            // Notify clients (e.g. an open recording session picker) that the
+            // set of online sessions changed, so they can refresh live.
+            this.io.emit("sessionsChanged");
```
dennis-zyska: This is sent to everyone, I think it is an security issue. Every client could track when sessions changes and can track statistics. We should not do this, only admins should be aware of this.

backend/webserver/Server.js
```diff
@@ -409,7 +422,57 @@ module.exports = class Server {
                         this.logger.warn("Failed to broadcast user monitor stats on disconnect: " + e);
                     }
 
+                    // If a recorded socket drops mid-recording, stop ONLY that
+                    // socket's recording and flag it "disconnected". Other active
+                    // recordings (including other admins' batches) keep running.
+                    try {
+                        const activeRecordings = this.activeRecordings || {};
```
dennis-zyska: can we move this functionility to the utils and just call one function here, otherwise this file becomes large and more and more unreadable

backend/webserver/Server.js
```diff
@@ -529,5 +596,30 @@ module.exports = class Server {
             this.logger.error("flushAllStats encountered an error: " + e);
         }
     }
+    /**
+     * Mark any recordings still in "recording" status as "interrupted".
+     * These are recordings whose server died mid-capture — the in-memory
+     * activeRecordings map is gone but the DB rows were never closed out.
+     * @returns {Promise<void>}
+     */
+    async recoverInterruptedRecordings() {
```
dennis-zyska: this should also be moved to a helper file in the utils folder

dennis-zyska (changes requested):
I did not reviewed all files yet, but already made some suggestions for improvement. Please have a look.

backend/webserver/sockets/recorder.js
```diff
@@ -0,0 +1,426 @@
...
+        return await snapshot(this.server.db.sequelize, this.logger);
+    }
+
+    /**
+     * Register this socket's event handlers.
+     * @returns {void}
+     */
+    init() {
+        this.createSocket("recorderStart", this.startRecording, {}, true);
+        this.createSocket("recorderStop", this.stopRecording, {}, false);
```
dennis-zyska: recorderStop is registered with useTransaction=false (the fourth parameter), but inside the function is an updateById.

backend/webserver/sockets/replayer.js
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    /**
+     * Group trace rows by socketId, falling back to "user-{userId}" when
+     * socketId is null (older recordings captured before per-session
+     * tracking was added).
+     * @param {Array<Object>} traces - Trace rows sorted by startTime
+     * @returns {Map<string, {userId: number, traces: Array<Object>}>} Map of session key to its traces and owning user
+     */
+    groupTracesBySocket(traces) {
```
dennis-zyska: to keep the sockets file clean, we put helper functions like groupTracesBySocket and buildActiveSessions into the helper files to ./backend/utils

frontend/src/components/dashboard/SocketProfiler.vue
```diff
@@ -0,0 +1,420 @@
...
+          },
+          title: "Delete recording",
+          action: "deleteRecording",
+        },
+      ],
+    };
+  },
+  computed: {
+    recordings() {
+      return this.$store.getters["table/recording/getAll"].map(r => ({
```
dennis-zyska: this.$store.getters["table/recording/getAll"] is called here, then again directly in exportRecording (line 231) and downloadReplayResults (line 353), instead of reusing the recordings computed. Looks like the direct calls are needed because recordings formats dates into display strings, so it can't be reused as is there. A separate rawRecordings computed that all three build on would avoid the repeated getter calls while keeping the raw values available where they're needed.

frontend/src/components/dashboard/socketprofiler/ImportRecordingModal.vue
```diff
@@ -0,0 +1,320 @@
...
+      let successCount = 0;
+      let failCount = 0;
+
+      // TODO: recording + traces are inserted as independent appDataUpdate calls
+      // with no shared transaction, so a server-side failure mid-loop leaves a
+      // partially imported recording. A dedicated backend import handler wrapping
+      // both in one transaction would make this atomic. Pre-validation above
+      // covers the common bad-file cases in the meantime.
+      // Re-create each trace under the new recording ID.
+      for (const trace of traces) {
```
dennis-zyska: Traces are imported one at a time via appDataUpdate in this loop, each awaited before the next starts, so an import of N traces costs N round trips. The TODO right above this already flags the atomicity gap (partial import on a mid-loop failure). A dedicated backend import handler wrapped in one transaction would fix both at once: one round trip instead of N, and the whole import succeeds or fails atomically.

dennis-zyska (changes requested):

backend/webserver/sockets/recorder.js
theultimatewasteofspace: Good catch set it to useTransaction=true. 
This also moved detachListeners() and the activeRecordings delete into afterCommit so a rollback can't leave the server out of sync with the DB, and the manual broadcast is now only used by the disconnect path, which has no transaction.

backend/webserver/sockets/replayer.js
theultimatewasteofspace: Moved both into backend/utils/replay-sessions.js and they're pure functions, so no behaviour change. And named it to match replay-worker.js and replay-auth.js in the same folder.

frontend/src/components/dashboard/SocketProfiler.vue
theultimatewasteofspace: Added a rawRecordings computed as the single store read so recordings now formats from it, and exportRecording and downloadReplayResults use it directly instead of calling the getter again.

frontend/src/components/dashboard/socketprofiler/ImportRecordingModal.vue
theultimatewasteofspace: Added a recorderImport handler with useTransaction=true and it creates the recording and all traces in one transaction, so the import is one round trip and either fully lands or fully rolls back. Validation is repeated server-side since the handler is reachable directly. Removed the TODO and the partial-import toast, neither applies now.

backend/webserver/sockets/recorder.js
```diff
@@ -0,0 +1,521 @@
...
+     * recording row, so one capture per participant.
+     * @param {Object} data - {participantSocketIds: string[], name?: string, excludeEvents?: string[]}
+     * @param {Object} options - Handler options; carries the transaction
+     * @returns {Promise<void>}
+     * @throws {Error} If the caller is not an admin, no sessions are selected,
+     *                 a session is already recording, or a session is offline
+     */
+    async startRecording(data, options) {
+        if (!(await this.isAdmin())) {
+            throw new Error("Admin access required");
```
dennis-zyska: Since #209, `Socket.js` only resolves a thrown error for the client if it's a `TranslatableError`, a `generateError()` shape, or a plain `Error` whose message *is* an i18n key. Everything else collapses to `errors.server.unexpectedError`.

So every plain-English `throw new Error(...)` in this file and `replayer.js` (~28 sites) now reaches the user as a generic "unexpected error" — the specific text never arrives. The validation ones hurt most: "One or more selected sessions are already being recorded" becomes unactionable.

The admin gates are a free fix — `errors.users.adminAccessRequired` already exists in `utils/modules/i18n/en/errors.json` with exactly this text, so all ~8 sites become `throw new TranslatableError("errors.users.adminAccessRequired")` with no new keys. The rest need keys under a new `errors.socketProfiler.*` namespace.

backend/webserver/sockets/replayer.js
```diff
@@ -0,0 +1,468 @@
...
+
+        // The scaling path runs level * N concurrent sockets at its peak, so
+        // MAX_CONCURRENCY (enforced per-level in runOneLevel) can be blown past
+        // here even when maxIterations is within MAX_ITERATIONS. Reject up front
+        // rather than open thousands of sockets against the server.
+        if (singleLevel === null) {
+            const peakConcurrency = maxIterations * pool.sessions.length;
+            if (peakConcurrency > MAX_CONCURRENCY) {
+                throw new Error(
+                    `Peak concurrency ${peakConcurrency} (${maxIterations} iterations x ${pool.sessions.length} sessions) exceeds the limit of ${MAX_CONCURRENCY}. Reduce --max-iterations or use fewer recordings.`
```
dennis-zyska: Interpolated messages need the params form: `new TranslatableError("errors.socketProfiler.peakConcurrencyExceeded", { peak, iterations, sessions, limit })`.

Unrelated to i18n but adjacent: this tells the user to "Reduce --max-iterations", but on the socket path the user is in the Start Replay modal, not the CLI. The flag advice only makes sense for the CLI caller.

backend/utils/recording-recovery.js
```diff
@@ -0,0 +1,70 @@
...
+
+        await recorder.stopRecording(
+            { socketId: socket.id, status: "disconnected" },
+            { internal: true }
+        );
+
+        const ownerSocket = server.io.sockets.sockets.get(entry.ownerSocketId);
+        if (ownerSocket) {
+            ownerSocket.emit("toast", {
+                title: "Recording stopped",
```
dennis-zyska: This emits a raw `{title, message}` toast straight to the wire, bypassing `sendToast`. Post-#209 `sendToast` takes i18n keys (see `sockets/study_session.js`), so this won't translate.

Same issue with the `sendToast` call in `replayer.js` `runScalingTest`, which passes an interpolated English string into what's now a key slot.

frontend/src/components/dashboard/SocketProfiler.vue
```diff
@@ -0,0 +1,432 @@
+<template>
+  <Card title="Socket Profiler">
```
dennis-zyska: The socket profiler UI is entirely hardcoded English — roughly 190 strings across this file and the nine `socketprofiler/*.vue` modals. Keys go in `utils/modules/i18n/{en,de}/`, both locales.

One structural note for when you do it: `tableColumns`/`tableButtons` are declared in `data()`, so they'd freeze at whatever locale was active on mount. The pattern on `dev` is to move them to `computed` so they re-evaluate on locale change — see `components/dashboard/Projects.vue`.

backend/db/migrations/20260506220447-extend-nav-socket-profiler.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const navElements = [
+  {
+    name: "Socket Profiler",
```
dennis-zyska, line 5: The English `name` here is right — `Sidebar.vue` uses it as the `$te` fallback — but the bundle entry is missing: there's no `sidebar.nav.socket_profiler` in `utils/modules/i18n/en/sidebar.json`, so the label stays English in German.

Separately: `path: "Socket_Profiler"` is capitalized where all 19 existing nav paths are lowercase (`admin_tools`, `session_overview`, …). `Sidebar.vue` derives the key as `sidebar.nav.${element.path}`, so this yields `sidebar.nav.Socket_Profiler`. Renaming isn't free though — `Topbar.vue`'s `onRecordingIconClick` hardcodes `/dashboard/Socket_Profiler`.

backend/db/models/recording.js
```diff
@@ -0,0 +1,65 @@
+"use strict";
+const MetaModel = require("../MetaModel.js");
+
+module.exports = (sequelize, DataTypes) => {
+    // Allowed recording lifecycle states. Single source of truth for the
+    // status validator below; add new states here rather than scattering
+    // string literals. "interrupted" is set on startup for recordings whose
+    // server died mid-capture.
+    const RECORDING_STATUSES = ["recording", "finished", "disconnected", "interrupted"];
```
dennis-zyska, line 9: These status values are rendered verbatim to users (Status column in `SocketProfiler.vue`, step 2 of the import modal), so `disconnected`/`interrupted` appear lowercase and untranslated. Wants a display-mapping layer — the stored values should stay as-is, since `recording-recovery.js` and `recorder.js` compare against them.

Same class of thing for the frontend-computed `PASSED`/`FAILED`, `Yes`, and `→ backend`/`→ frontend` labels in the results and traces modals.

dennis-zyska (changes requested):
We merged the i18n branch already, so a few things don't follow the convention anymore, please see the comments

backend/webserver/sockets/recorder.js
theultimatewasteofspace: Done. All 26 plain-English throws in recorder.js and replayer.js are now TranslatableError. The 9 admin gates use errors.users.adminAccessRequired, the rest are new errors.socketProfiler.* keys in en and de.

I left replay-worker.js:81 alone, it's a defensive check on a bad argument from an internal caller, not something a user can trigger, so the generic unexpected-error path fits it better. Say if you'd rather it were keyed.

Also switched the frontend toasts to resolveApiMessage(res, ...), since they were showing the pre-translated English message. German is my draft, maybe worth a check.

backend/webserver/sockets/replayer.js
theultimatewasteofspace: Changed to new TranslatableError("errors.socketProfiler.peakConcurrencyExceeded", { peak, iterations, sessions, limit }).

On the CLI advice, dropped the --max-iterations reference. The message now says to reduce the number of iterations or use fewer recordings, which reads correctly from both the modal and the CLI.

backend/utils/recording-recovery.js
theultimatewasteofspace: Fixed, it now emits titleKey/key instead of raw English, which Toast.vue resolves through $te/$t.

replayer.js runScalingTest is keyed too, but with one change worth flagging: sendToast(message, title, variant) has no params slot, so the interpolated level and reason couldn't be carried through. The message is now generic and points at the returned results. If you'd rather keep the values, sendToast would need a fourth params argument which is Socket.js change so I left it alone.

Separately, I added a notifyOnStop flag (default true) so the four perf CLI scripts skip that toast entirely since there's no browser to show it, and I had left a TODO in the code saying as much.

frontend/src/components/dashboard/SocketProfiler.vue
theultimatewasteofspace: Done. Parent plus all eight modals, 165 keys in a new utils/modules/i18n/{en,de}/socketProfiler.json, registered in both index.js files. Where a string already existed I reused it (common.cancel, common.close, common.add, common.createdAt) rather than duplicating. "ms" unit suffixes on latency values are left untranslated.
tableColumns/tableButtons moved to computed, same for the modal equivalents. 
oh and I count eight modals, not nine, pls let me know if I've missed one.

backend/db/migrations/20260506220447-extend-nav-socket-profiler.js
theultimatewasteofspace: Done. Path lowercased, Topbar.vue updated, sidebar.nav.socket_profiler added to both locales. Edited the migration in place since it's unmerged.

backend/db/models/recording.js
theultimatewasteofspace: Added a statusDisplay mapping through socketProfiler.status.*; stored values untouched so the backend comparisons still work. Import modal step 2 uses the same. PASSED/FAILED, Yes, and the direction labels are keyed too.

backend/tests/regression_stories/.gitattributes
```diff
@@ -0,0 +1 @@
+*.json linguist-generated=true
```
dennis-zyska, line 1: `030-upload-document.json` is 4 MB across 242k lines because its `documentAdd` payload inlines Shrek_Script.pdf as one JSON array element per byte, plus the full extracted script in `wholeText`. It costs about 16 JSON bytes per PDF byte, and it also puts a copyrighted screenplay into a public repo permanently. Could you re-record that story against a small project-owned PDF? Nothing after the `documentAdd` trace reads the document text, so the swap is safe, and a few-KB PDF brings the story under 100 KB. (Commenting here because GitHub rejects inline comments on the file itself: "diff is too large".)

backend/webserver/sockets/recorder.js
```diff
@@ -0,0 +1,553 @@
...
+     * @returns {void}
+     */
+    init() {
+        this.createSocket("recorderStart", this.startRecording, {}, true);
+        this.createSocket("recorderStop", this.stopRecording, {}, true);
+        this.createSocket("recordingGetTraces", this.getTraces, {}, false);
+        this.createSocket("recordingGetOnlineSessions", this.getOnlineSessions, {}, false);
+        this.createSocket("recordingGetPerfHealth", this.getPerfHealth, {}, false);
+        this.createSocket("recordingGetPerfStats", this.getPerfStats, {}, false);
+        this.createSocket("recorderImport", this.importRecording, {}, true);
```
dennis-zyska, line 586: Deleting a recording goes through the generic `appDataUpdate` path (`SocketProfiler.vue:422`, and the trace deletes in `RecordingModal.vue:218`), which runs no admin check. Every other handler registered here gates on `isAdmin()`, so delete ends up being the one destructive operation any logged-in user can trigger by passing a recording id. Could you add an admin-gated `recorderDelete` handler here and call it from the frontend, the same way `recorderImport` replaced the per-trace `appDataUpdate` loop?

dennis-zyska (changes requested):

backend/webserver/sockets/recorder.js
theultimatewasteofspace: Added recorderDelete, admin-gated and transactional, mirroring recorderImport. One handler covers both cases: with traceIds it deletes those traces, without it deletes the recording. That also collapses the per-trace forEach in RecordingModal into a single round trip. Verified as guest and it returns adminAccessRequired.

backend/tests/regression_stories/.gitattributes
theultimatewasteofspace: done

dennis-zyska (approved):

PR #222, opened 2026-05-22
Feat: budget management

MRawhani:
## Summary
Budget management service for AI cost enforcement, log status management, and concurrency control for chat requests.

## New User Features
- None (internal service module)

## New Dev Features
- New `budget.js` service module handling:
  - AI budget enforcement with two caps checked per request:
    1. Model global cap (`ai_model.costLimit`) : sum across all users
    2. Per-user share cap (`ai_model_share.costLimit`) - individual user spend
  - Each cap has its own `resetAt`
  - One in-flight request per (`userId`, `studySessionId`) - second request is rejected
  - Usage computed from `ai_log` on demand
  - In-flight `ai_log` rows have `NULL` cost and contribute $0 to sums
- New migrations for additional columns in `ai_model` and `ai_model_share` tables
- Updated schema definitions for both models

## Improvements
- None

## Bug Fixes
- None

## Known Limitations
- `chat.js` and `share.js` not yet updated to call the new budget functions (`beginRequest`, `cancel`, etc.)
- Bulk delete (wipe) of all shares when creating a new share not yet removed
- Reset and cancellation functionality not yet exposed in backend
- Study creation does not yet support creating a share with a cost limit

## Future Steps
- Adjust current service files (e.g. `chat.js`, `share.js`) to call the new budget functions such as `beginRequest`, `cancel`, etc.
- Remove the wipe (bulk delete) of all shares when creating a new share.
- Ensure params are fully prepared and parsed before being passed to `budget.js` functions ; preprocessing must happen within the calling service file; `budget.js` will save them as-is (see params doc in budget file).
- Treat `budget.js` as the sole responsible module for AI budget enforcement, AI log status management, and concurrency for chat requests.
- Add necessary new functionalities: e.g. add the cost limit for model in frontend and backend, allow study creation to include a share with a cost limit, add reset and cancellation functionality, and call budget service where needed.


backend/webserver/services/ai/budjet.js
akash9676, line 1: Name of the file is wrong.

akash9676 (changes requested):

PR #232, opened 2026-06-02
fix : skip setup wizard in dev-backend targets

mohammadsherif0:
Updates dev backend startup targets to skip the setup wizard by default, while keeping the wizard available through `make dev-wizard`.

## New User Features
None

## New Dev Features
None

## Improvements
- `make dev-backend` and `make dev-backend-watch` now set `DEV_SKIP_WIZARD=true` for faster regular dev startup.

## Bug Fixes
None

## Known Limitations
None

## Future Steps
N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ help:
 	@echo "make doc 							Build the documentation"
 	@echo "make dev-build       		  		Build the frontend (make dev-build-frontend) and run the backend in development mode"
 	@echo "make dev-backend      		 		Run backend in development mode"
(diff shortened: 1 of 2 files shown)


bingobongomann (commented):
Those two look totally fine, but we do have "make dev" and "make dev-wizard", so we can decide if we want the wizard. 
We should also add "make dev-backend-wizard" so the people developing on windows, where frontend and backend are started separately, can also run the wizard in dev mode. 
Otherwise this could only be done via WSL and "make dev-wizard"

bingobongomann (approved):
Looks good now!

Issue #230, opened 2026-06-01
[BUG] Email template saves without required placeholders on URL navigation


mohammadsherif0:
# Describe the Bug:  

When editing a template with a required placeholder, leaving the editor by changing the URL to the dashboard can save/persist the template without required placeholders. That template can still be selected in Settings and made public.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

# Steps to Reproduce the Behavior: 

(Describe as detailed as possible so others can reproduce the issue.) 

1. Open an email template in the template editor (type with required placeholders).
2. Remove or omit a required placeholder (e.g. ~assignmentName~) from the content.
3. Do not use the normal close/save flow; navigate away via the browser URL.
4. Open Settings and assign that template to a submission upload email setting or publish it.

# Expected Behavior:  

Leaving the editor without valid content should not persist incomplete template content. Templates missing required placeholders should not be assignable in Settings or publishable until validation passes.

# Actual Behavior:  

The template is saved without the required placeholders. It can be assigned in Settings and published. 

# System Information: 

- **CARE Software (Version/Branch):** feat-120-assigment_dashboard 

# Additional Information:  

Validation for required placeholders exists when the template editor is closed through the normal path; forced navigation via URL appears to bypass that check.

PR #235, opened 2026-06-03
Fix 230 missing required placeholder hardening

mohammadsherif0:
### Description
Fixes template editor leave behavior so invalid drafts (e.g. missing required placeholders) are not left behind or shown again after navigation away.

#230 

### New User Features
None
### New Dev Features
None
### Improvements
None
### Bug Fixes

- Warn before leaving the template editor when templateClose fails (missing required placeholders), with option to discard drafts and leave.
- Show a browser warning on tab close or forced URL when there are unsaved edits.
- Drop invalid template drafts on editor load so reopening no longer shows unmerged edits that omit required placeholders.
- Add templateDiscardDrafts socket to soft-delete draft template_edit rows without merging to template_content.

### Known Limitations
None
### Future Steps
N/A

### DIFF
--- a/frontend/src/components/Template.vue
+++ b/frontend/src/components/Template.vue
@@ -27,25 +27,47 @@
     subscribeTable: ["template"],
     components: {Loader, Editor},
     beforeRouteLeave(to, from, next) {
-      const templateEditor = this.$refs.editor?.$refs?.templateEditor;
-      if (!templateEditor || typeof templateEditor.requestClose !== "function") {
-        next();
-        return;
-      }
-      templateEditor.requestClose().then((res) => {
-        if (res && res.success) {
-          next();
-        } else {
-          if (res && !res.success) {
+      this.$nextTick(async () => {
+        const templateEditor = this.$refs.editor?.$refs?.templateEditor;
+        if (
+          !templateEditor
+          || typeof templateEditor.flushPendingEdits !== "function"
+          || typeof templateEditor.requestClose !== "function"
+          || typeof templateEditor.requestDiscard !== "function"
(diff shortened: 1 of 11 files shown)


frontend/src/components/editor/template/TemplateEditor.vue
```diff
@@ -307,6 +326,69 @@
...
+            (res) => resolve(res || { success: false })
+          );
+        });
+      },
+      /**
+       * Persist any pending debounced edits before close/discard checks.
+       * 
+       * @returns {Promise<void>}
+       */
+      flushPendingEdits() {
```
bingobongomann, line 353: This section could use some more comments/documentation

We do not have Placeholders and invalid updates in the other editors, but flushing of pending debounced edits would also make sense in the basic editor. Maybe we should move this function to the basic editor and use the flushing there too

frontend/src/components/Template.vue
```diff
@@ -27,25 +27,47 @@
...
+        if (
+          !templateEditor
+          || typeof templateEditor.flushPendingEdits !== "function"
+          || typeof templateEditor.requestClose !== "function"
+          || typeof templateEditor.requestDiscard !== "function"
+        ) {
+          next(false);
+          return;
+        }
+        try {
```
bingobongomann, line 45: Documentation about these hardening steps would be good, i did not see anything about it in the editor and templates docs.

bingobongomann (changes requested):

bingobongomann (approved):

dennis-zyska (approved):

Issue #192, opened 2026-05-02
[USER STORY] Prompt Templates


mohammadsherif0:
## Description

As a CARE user, I want prompt templates under Templates Dashboard that I can reuse when I use an LLM feature, with named placeholders filled from CARE before the model is called, and the template chosen at the point I start that action, so that I keep prompt text in one place and do not copy study, assessment, or comment data by hand.

### Placeholders

- `~pdfText~`: Text from the PDF in the current context.
- `~editorText~`: Text from the current editor document in the current context.
- `~assessmentResult~`: The saved rubric from the Assessment sidebar for this document and step (scores + notes).
- `~inlineComments~`: Structured comments/annotations for this document and step.
- `~nlpAssessmentSuggestion~`: The model’s draft rubric for this step (if the workflow ran NLP that pre-fills assessment). Empty if there is no suggestion.
- `~previousAssessmentResult~`: The saved rubric for the previous step, when this step is configured to reuse it. Empty if nothing is carried over.
- `~assessmentConfiguration~`: The assessment rubric configuration used in this step (criteria, scoring setup, and descriptions).
- `~submissionFiles~`: Primary submission document text plus a manifest of additional submission files.
- `~studyContext~`: Basic study/step/document metadata for the current context (for example study name, step name, and document title).

## Acceptance criteria

- [ ] A new prompt template type exists; users can create, edit, and manage these templates from Templates Dashboard.
- [ ] Templates support these named placeholders: `~pdfText~`, `~editorText~`, `~assessmentResult~`, `~inlineComments~`, `~nlpAssessmentSuggestion~`, `~previousAssessmentResult~`, `~assessmentConfiguration~`, `~submissionFiles~`, and `~studyContext~`.
- [ ] Missing placeholder values are replaced with empty text and the model is still called.
- [ ] Unexpected resolution failures stop before the LLM with a clear error.
- [ ] Unknown placeholders show a warning in the template editor.
- [ ] For each placeholder define one example value and verify that it resolves correctly.
- [ ] Everyone can create and edit prompt templates.
- [ ] The LLM feature gives the user the option to choose a prompt template in the same place as the feature (same screen or modal).
- [ ] Optional: Prompt template bodies support multiple languages like other templates (e.g. per-language content / selector), aligned with existing CARE template patterns.

## Estimated effort (Story Points)

TBD

## Tasks

- Add the prompt template type in Templates Dashboard.
[...]

PR #237, opened 2026-06-08
Feat 192 prompt templates

mohammadsherif0:
### Description

Adds Prompt templates (type 8) to the existing Templates system so authors can write AI prompts with study/document placeholders instead of hardcoded strings. Includes backend resolution via templateResolve, nine data-rich placeholders, and template-editor preview with example values.

#192 

### New User Features

- Create and edit Prompt templates from Dashboard → Templates (available to all users, not admin-only).
- Use the Placeholders sidebar to insert allowed prompt tokens (e.g. `~editorText~`, `~assessmentResult~`, `~nlpAssessmentSuggestion~`).
- Preview mode in the template editor substitutes placeholderExample values so authors can see how resolved prompts look before runtime use.

### New Dev Features

- New template type 8 (Prompt) with migration-backed placeholder rows and optional placeholderExample on the placeholder table.
- templateResolve socket resolves prompt placeholders from context + DB .
- buildPromptPlaceholderValues in backend/utils/templateResolver.js for type-8 resolution.
- backend/utils/studyNlpDocumentData.js for `~nlpAssessmentSuggestion~`.
- validateResolveContextAccess on templateResolve for document/session/step access checks.

### Improvements

Template editor placeholder preview extended to type 8 (alongside email types 1, 2, 3, 6, 7).

Bug Fixes

- Fix detached template refresh so copied templates update correctly after source changes.

### Known Limitations

- No study/AI UI caller for templateResolve yet, runtime wiring is expected with the AI integration.
- `~editorText~ `may need an explicit context.editorText override when resolving in the same pass as NLP insertIntoEditor; otherwise call after step loadingReady.
- Preview is template-editor only (example substitution), not live study-context preview for modal documents with chart placeholders.

Future Steps

- Wire templateResolve from the AI integration step flows once template selection UX is defined.
- Add modal-document preview for chart/comparison placeholders if needed for prompt authoring.


backend/utils/studyNlpDocumentData.js
```diff
@@ -0,0 +1,113 @@
+"use strict";
+
+/**
+ * Study-step NLP document_data key helpers.
+ *
+ * Keys match NlpRequest.saveResult: {service.name}_{service.skill}_{resultField}.
+ * Service discovery for assessment drafts matches Assessment.vue (nlpService /
+ * preprocessedAssessmentKeyCandidates).
+ */
```
akash9676, line 11: Author name

akash9676 (changes requested):

MRawhani (approved):
All is good. 
there are somethings that are against review Guidelines but I am not sure about how much these are met all over the code-base, for example
1- some functions don't have JSDocs 
2- the mapping between template types and their string labels (e.g. case 1: return "Email - General") is repeated and hard coded through multiple places, and when there is any update one has to go through them all.

Issue #159, opened 2026-04-14
[FEATURE] Export Grades


pdji1602003:
## Is your feature request related to a problem? Please describe 

This feature contributes to a bigger functionality where researchers are able to download project-specific datasets for subsequent analysis and research purposes.

## Describe the solution you'd like 

1. When clicking the `Export data` button on the `Projects` dashboard, the user can select `Export grades` and thus download the grades of selected students as a zip file. 
2. Throughout the downloading procedure, the user should be able to choose which format the downloaded grade file is, either in JSON or in CSV. 

## Describe alternatives you've considered 

N/A

## Additional context 

Related issues are:
1. https://github.com/UKPLab/CARE/issues/68
2. https://github.com/UKPLab/CARE/issues/151

pdji1602003:
Notes from the team meeting on 09.06:

1. In Export Data modal, add an option for CSV format: When the user selects CSV as the downloaded file format, offer an option that allows the user to merge CSV files. The merging logic goes like this:
    1.1 merge CSV files that belong to the same assessment configuration
    1.2 If there are different studies, merge the CSV files by study name and configurationId
    1.3 If there are different study steps, but share the same configurationId, name the CSV files like the following: studyName_stepId_configurationId (Add a column in the CSV to document the study step info)
2. Regarding the downloaded content, change "reviewer" to "sessionOwner".
3. Regarding the downloaded content, change "grader" to "studyOwner".
4. Regarding the downloaded content, remove roles.
5. Check if a normal user only sees user names of other sessions when they don't have the right of userPrivateInfo.
6. In StepOptions.vue, make sure to use BasicForm component to stay consistent in platform styling and to use statistics feature.

PR #238, opened 2026-06-09
Feat 68 export grades

pdji1602003:
Main Description

This pull request introduces grade exporting feature.

New User Features

- This feature allows the user to export project-specific grade data.
- When we are on CARE, which project we are in now is indicated by the button next to the user menu as shown in the below screenshot.
screenshot

-  This means that all the users, data we import from Moodle or operations we have done on CARE stay in this default project.
- Now when we want to download the project specific grade data, we can head to Projects > Manage > Export data.
- In the Export Data modal, choose the desired project for Project (in our case: Default Project) and then Export grades for Export Type. (Please make sure you already have students and submissions imported prior to this operation)
- Select students whose grades you would like to export (We can only export students who accepted data sharing).
- At Options step, choose whether you want to generate aliases for student names and grade file format (either CSV or JSON).
- At Confirm Download step, you could see a summary and then click on Download button.

Needs to Update according to https://github.com/UKPLab/CARE/issues/159#issuecomment-4659532944


backend/webserver/sockets/document.js
```diff
@@ -1109,12 +1112,14 @@ class DocumentSocket extends Socket {
             }
 
             let previousSubmissionId = null;
+            let resolvedProjectId = projectId;
```
melolw, line 1115: i don't understand what resolvedProjectId is?

frontend/src/components/dashboard/projects/export/StepOptions.vue
```diff
@@ -7,9 +7,9 @@
     <div class="form-check form-switch mb-2">
       <input 
         id="aliasSwitch" 
-        v-model="aliases"
```
melolw, line 10: don't forget to use basic form 🤣

melolw (approved):
i learned a lot of stuff that i was doing wrong by reading your code!! i couldnt find anything to comment honestly haha except for the basic form thing

frontend/src/components/dashboard/projects/export/StepOptions.vue
pdji1602003: Already updated it! :)

backend/webserver/sockets/document.js
pdji1602003: It's called resolvedProjectId because it assumes if the client side does not send a projectId for some reason, it would take the assignment.projectId as its value.

backend/webserver/routes/export.js
dennis-zyska, line 1: This file is growing to big, we should separate some helper functions into another file (e.g., sanitizeFolderName, ...)

dennis-zyska (approved):

PR #245, opened 2026-06-20
Sticky Stepper

mohammadsherif0:
To keep the step row visible while scrolling in the first-time setup wizard and in all StepperModal flows.

- SetupWizard: step indicator uses position: sticky with padding and a solid background so it stays pinned on long steps.
- StepperModal: same sticky treatment for the shared modal stepper, with a higher z-index so it stays above table search/headers when modal content scrolls.
- Connector line position updated to match step circle padding in both places.


bingobongomann (changes requested):
Image

In smaller screen resolutions there is a gap between the steps-topbar and the edge of the modal so we can see the scroll content behind the steps-topbar. 

This gap is caused by the padding of the model-body, but removing the upper padding to close that gap makes any non-stepper modal look weird, so we should add a conditioned scope for stepper modals, that sets the upper padding of the modal-body to 0, closing the gap and hiding the scrolling content behind it. 

Image

other than that it looks good and works as expected

dennis-zyska:
What is the status of this issue?

mohammadsherif0:
> What is the status of this issue?

Worked on the requested changes and waiting for final review of code.

bingobongomann (approved):
Modals look good now. Steps stay at the top without gap even in smaller resolutions

dennis-zyska (approved):

PR #178, opened 2026-04-22
fix 171 Wrong tooltip for toggle settings 

junaidferoz:
# Wrong tooltip for toggle settings

## What is the error?

The error was that the same tooltip option was being displayed on all the toggle settings in the settings tab. Which does not make sense.

## Where the error was.

This error is seen in `SettingItem.vue`, specifically here.

```
<div v-else-if="setting.type === 'boolean' || setting.type === 'bool'" class="form-check form-switch">
  <input 
    v-model="setting.value" 
    :checked="setting.value"
    class="form-check-input" 
    role="switch" 
    title="Activate/Deactivate NLP support"
    type="checkbox">
```

## The fix

Here, I changed the tile field to evaluate from the description of the settings. Please see the changes to get a clear understanding of what I did.

PR #247, opened 2026-06-23
Feat 176 single bulk assignment refactor

karimouf:
###Main Description
This pull request introduces a refactor to the single and bulk assignment

### New Dev Features

- Single assignments are no longer used
- Bulk assignment is now renamed to assignments
- The assignments model handles bulk and single assignment steps
- code visibly reduced
- Each step is in its own template component


frontend/src/components/dashboard/study/AssignmentModal.vue
```diff
@@ -0,0 +1,503 @@
...
+      <p class="text-center">Please create a study template to proceed!</p>
+    </template>
+
+    <!-- ─── Step 1: Template Selection ──────────────────────────────────────── -->
+    <template #step-1>
+      <TemplateSelectionStep
+          :key="`template-step-${stepResetKey}`"
+          ref="templateSelectionStep"
+          :modal-value="templateStepModalValue"
+          @update:modalValue="templateStepModalValue = $event"
```
pdji1602003: The event name should be hyphenated: @update:modal-value

pdji1602003:
[Question] Is the description field not mandatory? Without filling it out, I, as a user, can still submit the form.

sc1

pdji1602003:
[Bug] When creating a new assignment by clicking on "Add Assignment" button, the End Time of an assignment can be set before the Start Time.

pdji1602003:
[Bug] It's not in this pull request's scope😂, but I accidentally discovered if I click on the "Closed" tag, there would be this error "Uncaught TypeError: can't access property "closest", this._element is null". (All CARE platform shares this error.) 

sc2

Update: I am not sure if it has anything to do with this issue: https://github.com/UKPLab/CARE/pull/178

frontend/src/components/dashboard/study/assignment/AssignmentSelectionStep.vue
```diff
@@ -0,0 +1,282 @@
...
+      }
+    },
+    isValid() {
+      return this.bulk ? this.selectedAssignments.length > 0 : this.selectedAssignments.length === 1;
+    },
+  },
+  watch: {
+    selectedAssignments: {
+      handler(val) {
+        console.log('[AssignmentSelectionStep] selectedAssignments changed:', val);
```
pdji1602003: Just flag this console.log.

frontend/src/components/dashboard/Study.vue
```diff
@@ -67,8 +67,8 @@
     <ConfirmModal v-if="modals.confirm" ref="confirmModal" @hide="modals.confirm = false"/>
     <BulkCloseModal v-if="modals.bulkConfirm" ref="bulkConfirmModal" @hide="modals.bulkConfirm = false"/>
     <StudyCloseModal ref="studyCloseModal" />
-    <BulkAssignmentsModal v-if="modals.bulkAssignments" ref="bulkAssignmentsModal" @hide="modals.bulkAssignments = false"/>
-    <SingleAssignmentModal v-if="modals.singleAssignment" ref="singleAssignmentModal" @hide="modals.singleAssignment = false"/>
```
pdji1602003, line 71: [Question] Do we still need SingleAssignmentModal.vue since you've mentioned we no longer need it in the PR description?  If not, this file is still in the repo, do we need to remove it?

frontend/src/components/dashboard/study/assignment/WorkflowMappingStep.vue
```diff
@@ -0,0 +1,190 @@
...
+      this.$emit('update:isValid', val);
+    },
+  },
+  mounted() {
+    if (this.modalValue) {
+      if (this.modalValue.targetWorkflowId !== undefined) this.targetWorkflowId = this.modalValue.targetWorkflowId;
+      if (this.modalValue.workflowMapping) this.workflowMapping = { ...this.modalValue.workflowMapping };
+      if (this.modalValue.newStudyOwner) this.newStudyOwner = this.modalValue.newStudyOwner;
+    }
+    console.log('[WorkflowMappingStep] mounted isValid:', this.isValid);
```
pdji1602003: Just flag this so that you can remove it.

frontend/src/components/dashboard/study/assignment/WorkflowMappingStep.vue
```diff
@@ -0,0 +1,190 @@
...
+              s => s.id === currentSourceStep.workflowStepPrevious
+          );
+          if (previousSourceStep && previousSourceStep.stepType === 1) {
+            options.unshift({ name: '<Document> Revised Document', value: 'previousSubmission' });
+          }
+        }
+      }
+      return options;
+    },
+    getSubmission(studyId) {
```
pdji1602003: [Question] Where is this method used? I can't find it.

frontend/src/components/dashboard/study/assignment/TemplateSelectionStep.vue
```diff
@@ -0,0 +1,180 @@
...
+      deep: true,
+    },
+    assignmentTypeSelection: {
+      handler(val) {
+        this.$emit('update:modalValue', { templateSelection: this.templateSelection, assignmentTypeSelection: val, enableEmailNotification: this.enableEmailNotification });
+      },
+      deep: true,
+    },
+    assignmentType(val) {
+      this.$emit('update:assignmentType', val);
```
pdji1602003, line 135: There is no emit declaration, although there is emit event. (I think most of the step component or maybe all miss the emit declaration part, like this `emits: ['update:generateAliases', 'update:fakerSeed'],`

frontend/src/components/dashboard/study/assignment/TemplateSelectionStep.vue
```diff
@@ -0,0 +1,180 @@
...
+    isValid(val) {
+      this.$emit('update:isValid', val);
+    },
+  },
+  mounted() {
+    if (this.modalValue) {
+      if (this.modalValue.templateSelection) this.templateSelection = { ...this.modalValue.templateSelection };
+      if (this.modalValue.assignmentTypeSelection) this.assignmentTypeSelection = { ...this.modalValue.assignmentTypeSelection };
+      if (this.modalValue.enableEmailNotification !== undefined) this.enableEmailNotification = this.modalValue.enableEmailNotification;
+    }
```
pdji1602003, line 161: [Question] In other step components, you put isValid in mounted(), but you didn't in this component. Is there any reason?

frontend/src/components/dashboard/study/assignment/ConfirmationStep.vue
```diff
@@ -0,0 +1,189 @@
...
+    reviewerSelectionMode: { type: Object, required: false, default: () => ({}) },
+    workflowStepsAssignments: { type: Array, required: false, default: () => [] },
+    numberOfReviews: { type: Number, required: false, default: 0 },
+    reviewerSelectionModeFields: { type: Array, required: false, default: () => [] },
+    reviewer: { type: Array, required: false, default: () => [] },
+    documents: { type: Array, required: false, default: () => [] },
+    roles: { type: Array, required: false, default: () => [] },
+    roleSelection: { type: Object, required: false, default: () => ({}) },
+    reviewerSelection: { type: Object, required: false, default: () => ({}) },
+  },
```
pdji1602003, line 145: Maybe inject could be placed above props according to the official convention? 
Reference: https://vuejs.org/style-guide/rules-recommended.html

pdji1602003 (changes requested):
I just finished the review. Thank you for the hard work. I like how you use provide/inject for shared confirmation/distribution state to avoid deep prop drilling and also the idea of turning each step into an independent component for the separation of responsibility. :)

frontend/src/components/dashboard/study/assignment/ConfirmationStep.vue
```diff
@@ -0,0 +1,189 @@
...
+      <div><strong>Assignment Type:</strong> {{ assignmentType === 'document' ? 'Document' : 'Submission' }}</div>
+      <div v-if="assignmentType === 'document'">
+        <strong>Workflow Assignments:</strong>
+        <ul>
+          <li v-for="(stepAssignment, index) in workflowStepsAssignments[0]" :key="stepAssignment.workflowStepId">
+            - Workflow Step {{ index + 1 }}:
+            <span v-if="stepAssignment.documentId && documents.find(doc => doc.id === stepAssignment.documentId)">
+              {{ documents.find(doc => doc.id === stepAssignment.documentId).name }}
+              ({{ reviewer.find(u => u.id === documents.find(doc => doc.id === stepAssignment.documentId).userId)?.firstName }}
+               {{ reviewer.find(u => u.id === documents.find(doc => doc.id === stepAssignment.documentId).userId)?.lastName }})
```
NAO-200: Would it make sense to avoid repeating the same find() calls here?

frontend/src/components/dashboard/Study.vue
```diff
@@ -67,8 +67,8 @@
...
+    <AssignmentModal v-if="modals.bulkAssignments" ref="bulkAssignmentsModal" @hide="modals.bulkAssignments = false"/>
```
NAO-200: Would it make sense to pass :bulk="true" explicitly here as well? I think it would make the component usage a bit more self-explanatory

frontend/src/components/dashboard/study/assignment/DistributionStep.vue
```diff
@@ -0,0 +1,172 @@
...
+    };
+  },
+  computed: {
+    selectedReviewerRoles() {
+      return [...new Set(this.selectedReviewer.flatMap(obj => obj.roles))];
+    },
+    roleSelectionFields() {
+      return this.selectedReviewerRoles.map(roleId => ({
+        key: this.roles.find(role => role.id === roleId).id,
+        label: "Number of reviews for role: " + this.roles.find(role => role.id === roleId).name,
```
NAO-200: Would it make sense to resolve the role once (e.g. const role = this.roles.find(...)) and reuse it?

frontend/src/components/dashboard/study/assignment/ReviewerSelectionStep.vue
```diff
@@ -0,0 +1,168 @@
...
+    }
+  },
+  methods: {
+    userStudySessions(userId) {
+      return this.$store.getters["table/study_session/getFiltered"](s => s.userId === userId);
+    },
+    isStudyClosed(studyId) {
+      const study = this.$store.getters["table/study/get"](studyId);
+      if (!study) return false;
+      return study.closed === null;
```
NAO-200, line 156: I found study.closed === null a bit confusing. Is null the expected value for a closed study?

frontend/src/components/dashboard/study/assignment/WorkflowMappingStep.vue
NAO-200: I noticed that getSubmission() has the same implementation in AssignmentSelectionStep.vue as well. Would it make sense to extract it into a shared helper to avoid duplicating the logic?

NAO-200 (commented):
LGTM. I left a few minor comments regarding readability and maintainability, but overall the changes look good 👍

karimouf:
> [Question] Is the description field not mandatory? Without filling it out, I, as a user, can still submit the form.
> 
> sc1

No its not mandatory

frontend/src/components/dashboard/study/assignment/ReviewerSelectionStep.vue
karimouf: yes, the field closed inside the study table is not a boolean its a date

frontend/src/components/dashboard/study/assignment/ConfirmationStep.vue
karimouf: yes a function would be better not to repeat the code twice

frontend/src/components/dashboard/study/assignment/DistributionStep.vue
karimouf: yes that would be a bit cleaner

frontend/src/components/dashboard/Study.vue
karimouf: yes it makes sense to remove the file

frontend/src/components/dashboard/study/assignment/TemplateSelectionStep.vue
karimouf: there is no particular reason for that no

karimouf:
Thanks so much for your comments I found them very helpful I went through them and fixed all that required fixing please take another look @pdji1602003  @NAO-200

NAO-200 (commented):
LGTM

frontend/src/components/dashboard/study/AssignmentModal.vue
```diff
@@ -0,0 +1,503 @@
...
+
+    <!-- ─── Step 6: Bulk confirmation for session+bulk (only rendered for 6-step flow) ─── -->
+    <template #step-6>
+      <ConfirmationStep variant="session-bulk" />
+    </template>
+  </StepperModal>
+</template>
+
+<script>
+import StepperModal from "@/basic/modal/StepperModal.vue";
```
dennis-zyska, line 144: Docstring is missing

frontend/src/components/dashboard/study/AssignmentModal.vue
```diff
@@ -0,0 +1,503 @@
...
+    onError(res) {
+      this.$refs.assignmentStepper.stopProgress();
+      this.eventBus.emit("toast", {
+        title: "Failed to create assignment",
+        message: res.message,
+        variant: "danger",
+      });
+    },
+    createSingleAssignment() {
+      this.$refs.assignmentStepper.startProgress();
```
dennis-zyska, line 419: The progressId is not send here to the backend. Is it needed for single Assignment?

frontend/src/components/dashboard/Study.vue
dennis-zyska, line 470: why not having one function and send a parameter instead of the property like .open(bulk=true), it would reduce code and we only need to include the modal once. Better would be to have the selection even inside of the modal, in that way we only have once button here "Assignments"

frontend/src/components/dashboard/study/assignment/AssignmentSelectionStep.vue
```diff
@@ -0,0 +1,281 @@
...
+
+<script>
+import BasicTable from "@/basic/Table.vue";
+
+/**
+ * Step component for selecting the items to be assigned in the bulk assignment wizard.
+ * Renders a selectable table of documents, submissions, or study sessions depending
+ * on the assignment type chosen in the template step. Supports both single-select
+ * (for single assignment flow) and multi-select (for bulk flow).
+ * @author karim ouf
```
dennis-zyska: You moved a lot of code from the old files, where we had at least three authors of the code. Don't get rid of it and add them as well in **all new** files

dennis-zyska (changes requested):

frontend/src/components/dashboard/study/AssignmentModal.vue
karimouf: The progressId is used for tracking the progress of the bulk operation. In this case there is no bulk operation therefore we dont need to send the progressId to the backend

frontend/src/components/dashboard/study/AssignmentModal.vue
karimouf: I guess not its just down bellow

dennis-zyska (approved):

PR #255, opened 2026-07-01
[REFACTOR] Standardize buttons across the app

akash9676:
This update replaces plain HTML buttons with the shared `BasicButton` component across login, dashboard, editor, study, and annotation screens. 

### New User Features

- None

### New Dev Features

- ESLint rule blocks raw `<button>` tags in Vue files and directs developers to use `BasicButton`
- Exemptions for base button components so the rule stays practical

### Improvements

- Consistent button usage across auth, dashboard, editor, study, and PDF annotation views
- Standardized button attributes in authentication screens 
- Cleaner, more uniform button markup across 36 components
- ESLint compliance fix in the export format modal

### Bug Fixes

- None

### Known Limitations

- None

### Future Steps

- None


karimouf:
It looks fine, did you look at how all these changed buttons look now? Im afraid by this change some might appear off
additionally will you add each linting rule in a seperate PR?

frontend/src/auth/TwoFactorSelect.vue
```diff
@@ -56,24 +56,17 @@
...
             <div class="col-md-6 offset-md-4 my-4">
-              <button
+              <BasicButton
                 class="btn btn-primary btn-block"
                 type="submit"
                 :disabled="!selectedMethod || isSubmitting"
-              >
-                <span
-                  v-if="isSubmitting"
-                  class="spinner-border spinner-border-sm me-2"
```
pdji1602003, line 66: The loading state was removed because of the replacement from native button to BasicButton. Maybe we could expand the BasicButton here to accommodate this loading state by adding a new prop?

frontend/eslint.config.js
```diff
@@ -44,4 +44,26 @@ export default [
...
+                        'Use <BasicButton> from @/basic/Button.vue instead of raw <button>.',
+                },
+            ],
+        },
+    },
+    {
+        files: [
+            'src/basic/**/*.vue',
+            'src/components/**/Button.vue',
+        ],
```
pdji1602003, line 64: Do we have a clear rule as to what kind of stats data a shared component would send? If we exclude all the all basic components and shared components from the eslint rule, how do we know whether we will use native html button or BasicButton when we create a new shared component? Just raise this question for further discussion.

frontend/src/auth/TwoFactorSelect.vue
akash9676: I also thought of adding a loading state in the basic button. but i am not sure about the spinner though. 

@dennis-zyska what do you think?

dennis-zyska: yes, adding it with a property make sense

frontend/eslint.config.js
akash9676: we need to exclude them manually if we create any shared components with raw HTML

frontend/src/auth/TwoFactorSelect.vue
pdji1602003: Looks fine! Thanks for the change!

dennis-zyska (approved):

PR #262, opened 2026-07-11
Fix and refactor user related issue

karimouf:
## Summary

The user dashboard previously fetched data via a bespoke `userGetByRole` socket call and manual `fetchUsers()` refreshes wired into every child modal (`@update-user` events). It also broadcast full user records to every connected client regardless of role, which let non-admins read other users' `firstName`/`lastName`/`email`/`extId`.

This PR migrates the user dashboard onto the standard `appDataUpdate` + broadcast-table paradigm used elsewhere in the app, locks down exactly which columns get broadcast to whom, and fixes a `BasicTable` bug that would otherwise leave columns rendered with blank/missing data for users who aren't sent those fields.

## Changes

**Dashboard refactor**
- `Users.vue` now reads from `table/user/getAll` (the broadcasted table) instead of manually fetching via `userGetByRole` and re-fetching after every mutation. Removed `fetchUsers()` and the `@update-user` listeners on `DetailsModal`, `RightsManagementModal`, `RoleManagementModal`, `ImportModal`, and `UserCreateModal` — updates now arrive automatically through the table broadcast.

**Access control (backend)**
- Added a new right, `frontend.dashboard.studies.view.userPublicInfo`, scoped to `id` and `userName`, so every authenticated user (not just admins) receives a minimal public record for user-picker/attribution UI.
- Added the corresponding migrations granting `userPublicInfo` broadly (`extend-user-rights-public-info.js`, `extend-role-rights-public-info.js`).
- `Socket.js`: admins now always bypass column restriction and get every attribute; non-admins are restricted to the union of columns from their granted access rights. Previously the column restriction applied unconditionally (even to admins) and was assigned onto `allAttributes.include` — for Sequelize, `{ include }` *adds* attributes rather than restricting to them, so the filter was silently not working. Also fixed `getAutoTable`'s attribute list crashing when an access-map entry has no `columns`.

**`BasicTable` (generic fix)**
- Columns are now dropped from rendering when their `key` isn't present on any row in `data`, instead of rendering an empty/blank column. This means any table fed by a rights-filtered broadcast (not just Users) automatically hides columns the current user isn't authorized to see, rather than showing them empty.

## Open question for reviewers

`userPublicInfo` currently exposes just `id` and `userName`. Is that the right minimal set for cross-user visibility (e.g. for assignment/study collaborator pickers), or should anything else be added?

## Test plan
- [ ] Verify non-admin users see only `ID`/`User` columns in the Users table (First/Last Name columns absent, not blank)
- [ ] Verify admins still see all columns including First/Last Name/Email
- [ ] Create/edit/delete a user as admin and confirm the table updates live without a manual refresh
- [ ] Confirm no other `BasicTable` consumers regressed (columns that should show blanks per-row still do; only fully-absent keys are hidden)

### DIFF
--- a/backend/db/models/user.js
+++ b/backend/db/models/user.js
@@ -399,6 +399,7 @@ module.exports = (sequelize, DataTypes) => {
                 {
                     where: {id: userId},
                     returning: true,
+                    individualHooks: true,
                     transaction: options.transaction,
                 }
             );
--- a/backend/db/models/user.js
+++ b/backend/db/models/user.js
@@ -14,6 +14,10 @@ module.exports = (sequelize, DataTypes) => {
                 right: "frontend.dashboard.studies.view.userPrivateInfo",
                 columns: ["firstName", "lastName", "email", "extId"]
             },
+            {
+                right: "frontend.dashboard.studies.view.userPublicInfo",
+                columns:["id", "userName"]
+            }
         ];
 
         /**
--- a/backend/webserver/Socket.js
+++ b/backend/webserver/Socket.js
@@ -507,13 +507,14 @@ module.exports = class Socket {
             }
         }
 
-        // --- Column restrictions from access rights (applied regardless of row logic) ---
-        if (accessRights.length > 0) {
-            allAttributes['include'] = [...new Set(
-                accessRights
-                    .filter(a => a.columns)
-                    .flatMap(a => a.columns)
-            )];
+        // --- Column restrictions from access rights (admins bypass, everyone else is restricted) ---
+        if (!isAdmin && accessRights.length > 0) {
+            const columns = [...new Set(accessRights.filter(a => a.columns).flatMap(a => a.columns))];
+            if (columns.length > 0) {
+                // Pass a plain array so Sequelize restricts to only these columns.
+                // { exclude, include } is NOT the same — include there adds virtual attrs, not restricts.
+                allAttributes = columns;
+            }
         }
 
         // Apply row-visibility: baseFilter AND (condition1 OR condition2 OR ...)
@@ -838,7 +839,7 @@ module.exports = class Socket {
 
             let data = [];
             if (accessRights.length > 0 || await this.isAdmin(userId)) {
-                const attributes = [...new Set(accessRights.flatMap(a => a.columns))];
+                const attributes = [...new Set(accessRights.filter(a => a.columns).flatMap(a => a.columns))];
                 data = await this.models[table].getAutoTable(filter, userId, attributes);
             } else {
                 data = await this.models[table].getAutoTable(filter, userId);
--- a/frontend/src/components/dashboard/users/DetailsModal.vue
+++ b/frontend/src/components/dashboard/users/DetailsModal.vue
@@ -141,7 +141,7 @@ export default {
         roles,
       };
       this.$refs.modal.waiting = true;
-      this.$socket.emit("appDataUpdate", { table: "user", data: { id: userId, ...userData } }, (response) => {
+      this.$socket.emit("userUpdateDetails", { userId, userData }, (response) => {
         if (response.success) {
           this.$refs.modal.waiting = false;
           this.$refs.modal.close();
--- a/frontend/src/components/dashboard/users/DetailsModal.vue
+++ b/frontend/src/components/dashboard/users/DetailsModal.vue
@@ -57,7 +57,6 @@ export default {
   data() {
     return {
       userId: 0,
-      userInfo: {},
       formFields: [
         {
           key: "userName",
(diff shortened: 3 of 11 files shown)


backend/db/models/user.js
```diff
@@ -14,6 +14,10 @@ module.exports = (sequelize, DataTypes) => {
                 right: "frontend.dashboard.studies.view.userPrivateInfo",
                 columns: ["firstName", "lastName", "email", "extId"]
             },
+            {
+                right: "frontend.dashboard.studies.view.userPublicInfo",
+                columns:["id", "userName"]
```
akash9676, line 19: Here we only want to ID and User Name only but the GetAll has roles does it leaks the roles ?? as well instead of the id and username

backend/webserver/Socket.js
```diff
@@ -838,7 +839,7 @@ module.exports = class Socket {
 
             let data = [];
             if (accessRights.length > 0 || await this.isAdmin(userId)) {
-                const attributes = [...new Set(accessRights.flatMap(a => a.columns))];
+                const attributes = [...new Set(accessRights.filter(a => a.columns).flatMap(a => a.columns))];
```
akash9676, line 842: should we also do the same check !isAdmin && accessRights.length > 0 like this in the send table data?

frontend/src/components/dashboard/users/UserCreateModal.vue
```diff
@@ -117,15 +116,14 @@ export default {
       if (!this.$refs.form.validate()) return;
       this.$refs.modal.waiting = true;
 
-      this.$socket.emit("userCreate", this.formData, (response) => {
+      this.$socket.emit("appDataUpdate", { table: "user", data: this.formData }, (response) => {
```
akash9676: User create checks if the user is admin in app.js then we can create user but appupdatedata does not check for admin in app.js

frontend/src/components/dashboard/users/DetailsModal.vue
```diff
@@ -142,11 +141,10 @@ export default {
         roles,
       };
       this.$refs.modal.waiting = true;
-      this.$socket.emit("userUpdateDetails", {userId, userData}, (response) => {
+      this.$socket.emit("appDataUpdate", { table: "user", data: { id: userId, ...userData } }, (response) => {
```
akash9676: here also the appdata update skips the user check in app.js so we need to check it here or update the https://github.com/UKPLab/CARE/blob/c6113dca7a01b6bac4d924b054629978e41c54bb/backend/webserver/sockets/app.js#L65 funtion

akash9676 (changes requested):

frontend/src/components/dashboard/users/DetailsModal.vue
mohammadsherif0: Agree userUpdateDetails was admin-gated appDataUpdate is not.

frontend/src/components/dashboard/users/UserCreateModal.vue
mohammadsherif0: same concern here too

backend/db/models/user.js
karimouf: Roles have different compute for it that adds the roles of each user to isntance in the vuex store, it's implemented long ago

backend/webserver/Socket.js
karimouf: why would we need that there

frontend/src/components/dashboard/users/UserCreateModal.vue
karimouf: Nice catch, you are right, we now skip the admin check before we create the user, adding the check in the add function inside the user model file would fix the issue

frontend/src/components/dashboard/users/DetailsModal.vue
karimouf: You are also correct here. I looked at the differences between the two functions, and I guess that changing the function socket to the generalised one (appDataUpdate) is not the best option in this case. I will use the old socket call and make sure they are pass by the sequelize hooks

karimouf:
I would kindly request you take another look,
my changes are
1. changed the socket event for creating users back to its original function to make sure it checks for admin access
2. changed the socket event for editing user details back to its original function to make sure we check for admin access in addition to additional logic for user role mapping
3. I changed the user details get info which used an additional socket event to use the vuex store compute thought we dont really need the isAdmin guard in that scenario because we already omly send the full user information to the admin only 
@akash9676 @mohammadsherif0

backend/webserver/Socket.js
akash9676: i thought it is the same mechanism as send table data. If not no issue

dennis-zyska (approved):

PR #271, opened 2026-07-13
Feat 264 enhance sharing functionality

MRawhani:
## Summary
Replaces the service-command-based sharing system for AI models and hooks
with the standard subscribeTable + appDataUpdate pattern, soremoving the
custom `share.js` service file entirely.

## New Dev Features
- `MetaModel` gains a reusable `foreignOwner` config (`{column, table}`) plus a
  generic `validateForeignUserId`, letting any model accept a foreign `userId`
  (e.g. a share recipient) in `appDataUpdate` when the requester owns the
  referenced parent row. Any future share-like table needs only to declare its
  `foreignOwner` 
## Improvements
- Removed `backend/webserver/services/ai/share.js` (~630 lines) and its dead
  helpers (`userDisplayLabel`, `shareAggregatesFromRows`, `parseShareExpiryInput`,
  `mapShareToRecipient`). Only `getAiHookDisplaySummaries` (a redacted model-name
  lookup for hook recipients) had no subscription equivalent and was kept,
  relocated out of the deleted file.
- Frontend is rewritten to reflect that: the recipient picker now comes from
  `userGetByRole`, and share config is read/written through the Vuex store +
  `appDataUpdate` 

- `ai_model.js` / `ai_hook.js` using `parentTables`/`foreignTables` cascades so
  owner names and share/model-link rows propagate automatically through existing
  subscriptions.

## Bug Fixes
- Closed a gap where reactivating or soft-deleting an `ai_model_share` /
  `ai_hook_share` row via the `deleted`-field shortcut in `appDataUpdate`
  bypassed the foreign-owner check entirely (that shortcut path never reaches
  `validateForeignUserId`). Added a `beforeUpdate` guard (`validateOwnerUpdate`)
  directly on both share models, matching the pattern `ai_hook.js`/`ai_budget.js`
  already use for their own `beforeUpdate` hooks.
- Fixed role-based budget capping: `_findActiveShare` in
  `webserver/services/ai/budget.js` only matched a share row by direct `userId`,
  so a budget cap on a role-based share was never found/enforced for recipients
  who only had role access. It now also matches by the requester's role ids.

## Known Limitations
- The recipient picker's `userGetByRole` call only returns results for admins
  until the right `backend.socket.user.getUsers.all` is granted to non-admin
  roles (tracked separately, on another branch).
- beforeUpdate hook is a temp fix for a bug. A complete generic fix will be another issue


backend/db/models/ai_model_share.js
```diff
@@ -10,14 +10,58 @@ const {Op} = require("sequelize");
...
+                where: {id: aiModelId, deleted: false},
+                transaction,
+            });
+            return Boolean(model) && Number(model.userId) === Number(requesterId);
+        }
+
+        /**
+         * Guards direct updateById calls
+         */
+        static async validateOwnerUpdate(share, options = {}) {
```
karimouf, line 34: why is this needed? I get it's to guard updates, but isnt it already handled here https://github.com/UKPLab/CARE/pull/271/changes#diff-bebf1ced92fcdb11955cf946bd57d0585c51fc62de2b00637e14b0e67642c14fR39:~:text=static%20async-,validateForeignUserId,-(payload%2C

backend/db/models/ai_model.js
```diff
@@ -7,18 +7,47 @@
...
-                table: "ai_model_share",
-                by: "aiModelId",
-                target: "id",
-                columns: this.getAttributes(),
-            },
-        ];
+        // Cascade ai_model_share rows to anyone subscribing to ai_model, and the
+        // owner's user row back to anyone the model is shared with.
+        static autoTable = {
+            foreignTables: [
```
karimouf, line 17: dont get the foreign table and parent table parameter what do they add?

backend/db/models/ai_model_share.js
```diff
@@ -10,14 +10,58 @@ const {Op} = require("sequelize");
...
+         * Lets appDataUpdate accept a payload's userId as the share recipient (not the
+         * requester) when the requester owns the referenced ai_model. Called generically
+         * from AppSocket#updateData's foreign-userId guard.
+         *
+         * @param {{aiModelId?: number, id?: number}} payload Incoming appDataUpdate payload.
+         * @param {number} requesterId The socket's authenticated user id.
+         * @param {import("sequelize").Transaction} [transaction]
+         * @returns {Promise<boolean>}
+         */
+        static async validateForeignUserId(payload, requesterId, transaction) {
```
karimouf: I think this works fine overall, but it feels a bit redundant since nearly the same function is repeated in both `ai_model_share` and `ai_hook_share`.

I also find the name `validateForeignUserId` a bit misleading.

One way to avoid the duplication would be to define the function only once and make it configurable by setting the parent table and the foreign key used to fetch the parent record as static values in the model file. For example, in `ai_model_share`, the parent table would be `ai_model` and the foreign key would be `aiModelId`.

With those values defined statically, the same generic function could handle the validation for both models, eliminating the need to implement nearly identical logic in each model file.

backend/db/models/ai_model.js
MRawhani: populate values from other tables

backend/db/models/ai_model_share.js
MRawhani: from what i understood it is for updateById mostly. 

If you check app.js, there is a check for fields like "deleted, close ..." , this check doesnt go to the UserId check we were fixing. so it doesnt call that function.

before this https://github.com/UKPLab/CARE/pull/271/changes#diff-94068ba711c4d2080b3cc1cb6148d88f944c1d4cf3622f11dcec5bbac4d366abR87 in app.js

backend/db/models/ai_model_share.js
MRawhani: it can be, the idea is maybe validation might be different in other cases, if generalizing will be for all, then that is good.

Issue #275, opened 2026-07-18
[FEATURE] Configurable placeholder options for type 8 prompt templates


mohammadsherif0:
## Is your feature request related to a problem? Please describe 

For type 8 (prompt) templates, placeholders that inject large retrieved content, especially `pdfText` (and similarly `editorText` / submission file text), currently have no per-instance control over how much data is included. Size limits, if any, are applied globally in the resolver (e.g. a fixed character cap) rather than configured per placeholder usage. That makes it hard to tune prompts for model context limits, cost, and relevance without hardcoding behavior.

## Describe the solution you'd like 

Introduce a general mechanism for extra options on placeholders, starting with type 8 and the `pdfText` case, designed so it can be extended later for other options and other placeholders.

## Describe alternatives you've considered 

- Keep a single hardcoded global cap in `templateResolver` (current approach): simple, but not tunable per prompt or per placeholder instance.
- Hardcode limits only for specific placeholders in backend code: does not scale and still lacks per-template control.
- Remove all caps and allow unlimited retrieved content: avoids silent truncation, but risks oversized prompts, higher cost, and model context overflows with no per-usage control.

## Additional context 

- Tokens already use indexed form `~key[N]~`; options would extend that representation.
- Relevant UI: template sidebar / `TemplateConfigurator` when inserting placeholders: optional fields when adding a placeholder.
- Manual authoring: parse and validate curly-brace option syntax in the template editor; invalid or unknown options should fail validation clearly (or be ignored).
- Relevant resolve path: `backend/utils/templateResolver.js` (today uses a shared `capText` / `TEXT_PLACEHOLDER_CHAR_CAP` on retrieved/substituted placeholder content).
- This should be a **general placeholder-options feature**, with char limit for `pdfText` as the first concrete use case, not a one-off special case only for PDF text.
- Scope note: this is about limiting **retrieved/substituted placeholder content**, not about limiting text authored in the template editor itself.

Issue #248, opened 2026-06-23
[FEATURE] Placeholder order and numbering extension for prompt templates


mohammadsherif0:
## Is your feature request related to a problem? Please describe 

- Prompt templates (type 8) can include the same placeholder multiple times (e.g. `~submissionFiles~`, `~pdfText~`). Today every instance resolves to the same value, so authors cannot target a specific file or ordered item per slot.
- `submissionFiles` naming/description is also unclear about ordered/multi-file use.

## Describe the solution you'd like 

- Extend placeholder resolution with **order/numbering** (array-style), so each numbered instance resolves to the item at that index (e.g. 1st PDF, 2nd PDF).
- Support differentiating multiple occurrences of the same placeholder in one prompt.
- Update `submissionFiles` placeholder **wording** (label/description) to reflect ordered/multi-file behavior.
- Unnumbered placeholders keep current behavior (same value everywhere, or sensible default for index 0).

## Describe alternatives you've considered 

- Separate placeholder keys per file (e.g. `submissionFile1`, `submissionFile2`) — rejected.

## Additional context 

None

PR #276, opened 2026-07-20
Feat 275 placeholder options

mohammadsherif0:
## Summary
Prompt (type 8) placeholders can carry per-token `wordRange` and `pageRange` options so authors limit what text is sent to the model. Limits are applied when the prompt is resolved, from PDF.js page lists at hook extract time, not as a second limiter in the hook InputMap.

## New User Features
- Type 8 placeholders `pdfText` and `submissionFiles` can take Word range and Page range; `editorText` is seeded with Word range only.
- The Placeholders sidebar collects From/To before insert, for example `~pdfText[1]{pageRange:1}~` or `~editorText[1]{wordRange:1-500}~`.
- Word range and page range can be set on the same token. Pages are applied first, then words (`{pageRange:1,wordRange:1-200}` is the first 200 words of page 1).
- Omit options to send the full extract. Unknown or invalid option values are listed in the existing invalid-placeholder warning.

## New Dev Features
- Shared `utils/modules/placeholder-tokens` parses option braces, validates ranges, and applies `applyTextRangeLimit` (pages then words). Covered by unit tests including reversed option-key order, overflow clamp, and `{ pages, pageCount }` replacement (not `[object Object]`).
- `placeholder.placeholderOptions` holds option definitions; migration seeds `wordRange` / `pageRange` on the type 8 text placeholders.
- Hook PDF extract returns `{ pages, pageCount }` from PDF.js `numPages`. `resolveServiceInput` keeps that object when the mapping is `pdf` so the resolver can slice pages.
- Saving a template rejects a repeated option name on one token (e.g. two `wordRange` values).

## Improvements
- Silent 2000-character caps on `editorText` and `submissionFiles` are removed; omit options to send the full extract.
- Frontend and backend consume the shared placeholder-tokens module instead of duplicated token helpers.
- Developer docs cover option syntax, apply order, TeX/zip as one implicit page, and where type 8 resolve runs (`hook.js` vs `templateResolve`).

## Bug Fixes
- Template editor preview state is declared in `data` so preview mode does not fail on first use.

## Known Limitations
- A single number `N` always means the first N units, not unit N only. `pageRange:2` is pages 1–2; page 2 only is `pageRange:2-2`.
- Page slicing needs a `pages` array from PDF.js at hook extract. Backend-only `templateResolve` without that object treats the extract as one page. TeX/zip `submissionFiles` text is also one implicit page.
- Extend placeholderOptions and resolver handling to other placeholders and additional option types beyond limits.

#275 
#248


frontend/src/components/editor/sidebar/TemplateConfigurator.vue
```diff
@@ -289,20 +609,36 @@
...
           templateId: this.templateId,
-          text: formatPlaceholderToken(placeholder.id, nextIndex),
+          text: formatPlaceholderToken(placeholder.id, nextIndex, optionsToInsert),
         });
+
+        if (this.hasPlaceholderOptions(placeholder)) {
+          this.pendingOptionRowsByKey[placeholder.id] = [];
+        }
       },
       getPlaceholderIcon(placeholderType) {
```
akash9676, line 594: No comment for this function

frontend/src/components/editor/sidebar/TemplateConfigurator.vue
```diff
@@ -199,6 +281,198 @@
       this.eventBus.off("editorContentUpdated", this.editorContentHandler);
     },
     methods: {
+      /**
```
akash9676, line 286: From there to line to like 475 there are many funtions that are like very small do we need that many funtions?

frontend/src/components/editor/template/TemplateEditor.vue
akash9676, line 1: This file is very Big you can split it into multiple files for better redability

backend/utils/helper/templateResolver.js
akash9676, line 1: This File is also very Huge you can split it into multiple files

akash9676 (changes requested):

frontend/src/components/editor/sidebar/TemplateConfigurator.vue
mohammadsherif0: They are needed. The sidebar template calls them per placeholder/row.

frontend/src/components/editor/template/TemplateEditor.vue
mohammadsherif0: This PR only adds previewMode to data. Splitting the editor here would mix a refactor into the feature. I can open a follow-up issue later if a clean split exists.

Issue #272, opened 2026-07-13
[FEATURE] extract shared dashboard patterns into basicDashboard foundation


mohammadsherif0:
## Is your feature request related to a problem? Please describe 

Dashboard UI under `frontend/src/components/dashboard/` grew page-by-page. The same patterns (card shell, table options, row actions, delete confirm, import/export toolbar, badges) are copy-pasted with small inconsistencies in icons, button colors, and badge semantics. That makes new dashboard pages slower to build and harder to keep visually aligned.

### Goal
Audit the full dashboard tree, classify what is shared vs page-specific, then extract a **`basicDashboard` foundation** so new pages are easier to implement and existing ones can migrate incrementally without breaking domain logic.

### Classification
| Layer | Meaning | Action |
|-------|---------|--------|
| **Redundant** | Duplicate/dead/conflicting code | Remove or merge first |
| **Universal** | Same on most list dashboards | Extract to `basicDashboard` base |
| **Common** | Many pages, not all | Shared helpers or optional building blocks |
| **Some** | 2–4 related pages | Domain subfolder shared component |
| **Specialized** | One page or heavy domain logic | Document and keep local so do not generalize |

### Phase 1: Shared primitives

- Default `BasicTable` options constant (plus optional search variant)
- Table height constants for full pages vs modals
- Thin delete helper: confirm dialog + soft delete via `appDataUpdate`
- Migrate `BasicCoordinator` footer buttons to `BasicButton`
- One toast pattern in files already being touched
**Standards to apply when touching row actions and badges:**
- Delete: `trash` + `btn-outline-danger`
- Edit, copy, open, download: one icon and style per semantic action
- Disabled assignment badge must not use green
`BasicCoordinator` stays optional. Complex flows keep custom modals.

### Phase 2: List page component

Add `DashboardListPage` (name flexible): card shell, header toolbar slot, `BasicTable` wired to Phase 1 defaults.
Slots: header actions, modals, table extras.
Migrate list-style pages incrementally. Good first candidates: Documents, Projects, Tags, Assignments, Templates, Workflows, Users, Configurations, NLP Skills, Logs.

### Phase 3: Follow-up

- Accordion layout for Submissions and Study Sessions
- Domain helpers (assignment/submission status, study operations) after Phase 2 is done

### Keep specialized
Studies, Settings, Configurations, Workflows graph editor, project full export, NLP service health UI, Logs server pagination, User Statistics dual tables, Session Overview custom filters.

[...]

Issue #268, opened 2026-07-13
[BUG] canceling project delete still resets default project


mohammadsherif0:
# Describe the Bug:  

Starting a project delete resets the default project even if the user cancels.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

# Steps to Reproduce the Behavior: 

1. Set default project to something other than project 1.
2. Dashboard → Projects → delete a different project.
3. Click Cancel on the confirm dialog.
4. Check which project is default.

# Expected Behavior:  

Default project stays unchanged when delete is cancelled.

# Actual Behavior:  

Default is reset to project 1.

PR #301, opened 2026-07-27
Feat 272 dashboard refactor

mohammadsherif0:
## Summary
Adds a shared dashboard list-page foundation (`DashboardListPage`, table defaults, row-action catalog, `confirmSoftDelete`) and migrates the card+table management screens onto it. Issue #272 list-page work is in this PR; remaining specialized layouts stay as they are.

## New Dev Features
- `DashboardListPage` (`frontend/src/basic/dashboard/ListPage.vue`): card + header slot + `BasicTable`, default options and height from `constants.js`.
- Row-action catalog, `DASHBOARD_BADGES`, and `confirmSoftDelete` in `actions.js` (confirm + `appDataUpdate` with `deleted: true`).
- Contributor recipe in `docs/source/for_developers/frontend/components/dashboard.rst`, component reference in `basic/dashboard.rst`, and the table example updated to use `DashboardListPage`.

## Improvements
- List-style pages now use `DashboardListPage`: Documents, Projects, Tags, Assignments, Templates, Workflows, Users, Configurations, NLP Skills.
- Nested tables that already had a manage column use the same catalog: `StudySessionTable`, `AssignmentSubmissionsTable`, `SavedTemplatesModal`, `PublicTemplatesModal`.
- Studies stays a custom Card/table shell but uses the catalog, `withSearch`, and `confirmSoftDelete`.
- `BasicCoordinator` and `ConfirmModal` footers use `BasicButton`.
- Tags/Projects share buttons use `filterMode: "and"` so Share only shows on a private row you own.

## Bug Fixes
- Deleting a project no longer resets `projects.default` if the user cancels the confirm dialog. #268 

## Known Limitations
These differ from issue #272 on purpose:

- **Logs** were listed as both a migrate candidate and specialized (server-side pagination). They stay a custom Card + `BasicTable` so pagination is not forced through `DashboardListPage`.
- **Configurations** were listed as both migrate and specialized. The list shell migrated; JSON viewer / import-export domain logic stayed on the page.
- **Studies** stay specialized (no `DashboardListPage` wrap). Catalog alignment was applied because that page still has a manage column.
- Issue asked for a **modals slot** on the list page. Shipped pages put modals as siblings of `DashboardListPage` so refs and lazy `v-if` stay on the page. `#headerActions` and `#afterTable` are the slots in use (`#afterTable` on Documents). `#tableExtras` exists for BasicTable extra toolbar buttons and has no list-page caller yet.
- Issue asked for **page vs modal table height** constants. Only `DASHBOARD_TABLE_HEIGHT` (`65vh`) is in the foundation. Modal tables in the repo do not share one height, so there is no modal-height constant.
- **Templates** and **assignment submissions** still delete via `templateDelete` / `submissionDelete`, not `confirmSoftDelete`, because those are dedicated backend sockets.
- Submissions and Study Sessions keep their accordion shells. Only the nested tables got catalog buttons.

## Future Steps
Further generalization and refactors can be planned later where the same pattern is still copied (remaining dashboard layouts, other frontend shells, and similar duplication elsewhere) so the rest of the tree can align with this kind of shared foundation.


frontend/src/basic/dashboard/constants.js
```diff
@@ -0,0 +1,35 @@
...
+  pagination: 10,
+});
+
+/**
+ * Same as the default options, but with search turned on.
+ *
+ * @param {Object} [options] - Options to start from
+ * @returns {Object} A new options object with search: true
+ */
+export function withSearch(options = DEFAULT_DASHBOARD_TABLE_OPTIONS) {
```
dennis-zyska: `withSearch({ sort: ... })` replaces the defaults instead of adding to them, so `pagination: 10`, `striped` and `hover` are dropped. Users.vue and Workflows.vue both call it that way and only still work because `ListPage.vue` merges the defaults back in. Make it safe on its own: `return { ...DEFAULT_DASHBOARD_TABLE_OPTIONS, ...options, search: true };`

frontend/src/components/dashboard/Study.vue
```diff
@@ -612,36 +542,31 @@ export default {
...
-                    variant: "success"
-                  });
-                } else {
-                  this.eventBus.emit('toast', {
-                    title: "Study delete failed",
-                    message: result.message,
-                    variant: "danger"
-                  });
-                }
+      this.modals.deleteConf = true;
```
dennis-zyska: These two lines copy the body of `openDeleteConfModal` (line 389), which was this method's only caller, so that helper is now dead code. Call the helper instead and pass the `confirmSoftDelete` callback to it. That also brings back the `?.` guard on `this.$refs.deleteConf` that this version drops.

frontend/src/components/dashboard/Users.vue
```diff
@@ -402,34 +357,29 @@ export default {
...
-                  this.eventBus.emit("toast", {
-                    title: "User not deleted",
-                    message: result.message,
-                    variant: "danger",
-                  });
-                }
-              }
-          );
-        }
+      this.modals.confirm = true;
```
dennis-zyska: Same as in Study.vue: these two lines copy the body of `openConfirmModal` (line 316), which was its only caller, so that helper is now dead code. Call the helper instead and pass the `confirmSoftDelete` callback to it.

frontend/src/basic/modal/ConfirmModal.vue
```diff
@@ -17,26 +17,23 @@
       </div>
     </template>
     <template #footer>
-      <button
+      <BasicButton
         class="btn btn-secondary"
-        type="button"
+        text="Abort"
```
dennis-zyska: On `dev` this button already reads `$t('common.abort')`, so merging this branch would replace the translation with hardcoded English. Please run `git merge dev` first, then use `:title="$t('common.abort')"` here (and `common.confirm` below). Passing `title` rather than `text` also matters because BasicButton sends a `stats` event on click, and with `title` unset that event carries an empty payload.

dennis-zyska (changes requested):

dennis-zyska (approved):

PR #309, opened 2026-07-29
fix: add unique index on users that are  not deleted

karimouf:
## New dev features
- Users can sign up again (or be re-invited) using an email, username, ORCID iD, or SAML NameID that belonged to a previously deleted account, instead of being blocked by a stale record

---

## Improvements
- Uniqueness on `user.email`, `userName`, `extId`, `orcidId`, and `samlNameId` is now scoped to active (`deleted = false`) rows via a partial unique index, matching the existing app-level logic (`filterExistingEmails`) instead of contradicting it

---

## Bug fixes
- **Fixed:** creating or bulk-importing a user with an email, username, `extId`, `orcidId`, or `samlNameId` that belonged to a soft-deleted account failed with `SequelizeUniqueConstraintError`, even though the app treated that value as available

---

## Known limitations
- Multiple soft-deleted users can now legitimately share the same email or username — any report or admin query that scans all rows (not just `deleted = false`) must not assume email or username is globally unique
- The migration's rollback (`down`) re-adds a plain unique constraint and will fail if an active + deleted row pair with a duplicate value exists at that point

---

## Future steps
- Automate this pattern: any model with a unique field should get a partial unique index (`WHERE deleted = false`) by default, rather than each column needing its own one-off migration — e.g. a shared migration helper or a lint/CI check that flags a plain `unique: true` on a column in a table that also has a `deleted` column


dennis-zyska (changes requested):
We should add a huge number of columns for that, have a look here: https://github.com/UKPLab/CARE/blob/dev/backend/db/migrations/20251005114112-transform-document_data.js

dennis-zyska (approved):

PR #314, opened 2026-07-31
fix : enable first-time setup wizard on main production build

mohammadsherif0:
Don't skip wizard for make ENV=main build

### DIFF
--- a/.env.main
+++ b/.env.main
@@ -15,8 +15,9 @@ ADMIN_PWD=admincare
 GUEST_EMAIL=guest@guest.com
 GUEST_PWD=guestguest
 
-# Skip first-time setup wizard: create admin from ADMIN_EMAIL / ADMIN_PWD and mark wizard complete (demo / dev pipelines).
-DEV_SKIP_WIZARD=true
+# First-time setup wizard: keep false for production/main builds so fresh installs use /wizard.
+# Dev pipelines use .env.dev or make dev (DEV_SKIP_WIZARD=true).
+DEV_SKIP_WIZARD=false
 
 # nlp server
 SERVICE_NLP_ENABLED=true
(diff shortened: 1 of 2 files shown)


.env.main
```diff
@@ -15,8 +15,9 @@ ADMIN_PWD=admincare
 GUEST_EMAIL=guest@guest.com
 GUEST_PWD=guestguest
 
-# Skip first-time setup wizard: create admin from ADMIN_EMAIL / ADMIN_PWD and mark wizard complete (demo / dev pipelines).
-DEV_SKIP_WIZARD=true
+# First-time setup wizard: keep false for production/main builds so fresh installs use /wizard.
+# Dev pipelines use .env.dev or make dev (DEV_SKIP_WIZARD=true).
```
dennis-zyska: This is not true, we have a dev and a main instance in the pipeline, both are published on our server. The .env.dev doesn't mean it is only used in make dev.

Dockerfile
```diff
@@ -23,4 +23,4 @@ WORKDIR content-server
 RUN echo $ENV
 RUN make ENV=$ENV build-frontend
 
-CMD ["sh", "-c", "make ENV=$ENV NODE_ENV=production init dev-backend"]
+CMD ["sh", "-c", "make ENV=$ENV NODE_ENV=production init dev-backend-wizard"]
```
dennis-zyska, line 26: Just to make sure, it is only used in the first run until settings are set?

dennis-zyska (changes requested):

Dockerfile
mohammadsherif0: Yes the wizard only runs while app.setup.wizardCompleted is unset/false. After setup completes, that setting is stored in the DB and the app routes to login/normal use instead of /wizard.

I also tested it locally and it works as expected.

dennis-zyska (approved):

PR #320, opened 2026-08-02
Feat 331 data export

melolw:
## Summary

Adds a new data-export flow to the project dashboard, streaming zip archives directly from the backend instead of building them client-side. Extends the existing submissions/grades exports with new **documents**, **studies** and **user behaviour** export type, and reworks the shared export UI (user selection, per-type options, confirmation) to support it alongside the existing types.

## Added Features

**New export type: User Behaviour**
- Streams `statistic` table rows (site usage/interaction logs) for selected users into the export archive, admin-gated (`resolveIsAdmin`).
- Choice of output layout: a single combined file, or one file per user (`StepOptionsUserBehaviour.vue`).
- Choice of file format: JSON (nested `data` object) or CSV (flattened `data` field) — same underlying `action, data, timestamp, user, username, userId, session` shape either way.
- Backend streams rows via keyset pagination (`id`-based, not `OFFSET`) directly into the archive using custom `Readable` streams (`createJsonArrayStream`/`createCsvRowsStream` in `utils/helper/export.js`), so memory use stays bounded regardless of table size — the naive full-table-dump approach (mirroring the existing `statsGet` socket handler) hit `RangeError: Invalid string length` on a 2M-row table during testing.
- User-selection table for this type lists every user on the platform rather than reusing project-scoped submission/document counts, with an "Accept Behaviour Sharing" column (`user.acceptStats`) instead of a count.

**Studies export**
- New "Include AI-assisted scores" option, tagging grade records whose source step configuration includes an `nlpRequest` service, split into a separate `scores_ai.json` per session (alongside `scores.json` for non-AI-assisted grades) — optional, default on.
- Workflow filter replaced with a dropdown + checkboxes (Select All / individual workflows), scroll-capped, instead of an inline checkbox list.
- "Include grades" renamed to "Include scores" (`scores.json` instead of `grades.json`) for consistency with the new AI-scores split.

**Documents export**
- Document-type checkboxes now include a compact "Select All" / "Unselect All" text-link toggle next to the field label instead of separate buttons.

**Shared user-selection table (`StepSelectUsers.vue`, formerly `StepSelectStudents.vue`)**
- Generalized to work across all export types via a shared `getOrCreateRow`/`forEachUserRow` helper, removing per-type duplication.
- New "Roles" column (from `user_role`/`user_role_matching`) across all export types, following the same pattern as the grades "Assessment Configuration(s)" column: compact numeric IDs in the cell, filter dropdown resolves them to readable `"id: name"` labels (fixes filtering being broken for multi-value cells, which previously matched on a single concatenated string like `"Multiple (A, B)"`).
- "Last Submitted" column removed entirely (and its backing per-row date tracking).
- Selection state now resets when the export type or project changes (`ExportModal.vue`'s new `resetOptions()`), fixing stale/mismatched selection counts (e.g. "174/173 selected") when switching between export types with different eligible-user sets.

**Grades export**
- Annotations now carry a `tagName` field (resolved from `tagId`) alongside the existing raw ID, inserted immediately after `tagId` in each record.


backend/webserver/routes/export.js
```diff
@@ -775,22 +394,557 @@ module.exports = function (server) {
...
+    async function processDocumentBasedExport(server, projectId, userIds, documentTypes, shouldExcludeNonConsentingEdits, shouldExcludeNonConsentingAnnotations, baseFolderName, archive) {
+        try {
+            documentTypes = typeof documentTypes === 'string' ? JSON.parse(documentTypes) : documentTypes;
+            if (!Array.isArray(documentTypes)) documentTypes = [0, 1, 2, 4];
+        } catch (e) {
+            console.warn("Could not parse documentTypes:", documentTypes);
+            documentTypes = [0, 1, 2, 4];
+        }
+        
+        const docs = await server.db.models.document.findAll({
```
dennis-zyska, line 557: Neither `processDocumentBasedExport` nor `processStudyBasedExport` verifies that the requesting user (`currentUserId`) actually has access to `projectId`, the route only checks that the project row exists (`loadExportRequestContext`), not that the caller is a member/owner/has rights on it. Any authenticated user can POST an arbitrary `projectId` + `userIds` here and receive that project's documents, annotations, comments, document edits, and study session data in the ZIP. 

The same gap applies to `processStudyBasedExport` (routes/export.js:608, `server.db.models.study.findAll({ where: studyWhere })`).

frontend/src/components/dashboard/projects/export/StepConfirmDownload.vue
```diff
@@ -63,15 +63,40 @@ export default {
...
     }
   },
   computed: {
     hasDeclinedSharingSelected() {
-      return this.submissionSelection.some(row => row.acceptDataSharing === 'No');
-    }
+      return this.userSelection.some(row => row.acceptDataSharing === 'No');
+    },
+    exportTypeLabel() {
+      const labels = {
```
dennis-zyska, line 80: What about 'userBehaviour'? Can we do it dynamic?

frontend/src/components/dashboard/projects/export/StepSelectUsers.vue
```diff
@@ -0,0 +1,273 @@
...
+          lastSubmissionDate: null,
+          configurationName: this.gradeConfigurations[uid]?.ids?.length
+            ? this.gradeConfigurations[uid].ids.join(", ")
+            : "No configuration",
+          configurationIds: this.gradeConfigurations[uid]?.ids ?? [],
+        };
+        return submissionsByUser[uid];
+      };
+
+      if (this.exportType === 'submissions' || this.exportType === 'userBehaviour') {
```
dennis-zyska: I'm not sure if we can reuse the submission export type here. The count in `getExportTypeColumn` would be wrong as it is based of the user's submission documents.

backend/utils/helper/export.js
```diff
@@ -0,0 +1,735 @@
...
+    return new Readable({
+        read() {
+            if (fetching || finished) return;
+            fetching = true;
+
+            (async () => {
+                try {
+                    const rows = await fetchPage(lastId, pageSize);
+                    if (rows.length === 0) {
+                        if (!started) this.push(Papa.unparse([mapRow].length ? [] : []));
```
dennis-zyska, line 683: If the userBehaviour CSV export has zero rows, the stream would push an empty string instead of the CSV header and empty rows. Just use Papa.unparse([])

backend/utils/helper/export.js
```diff
@@ -0,0 +1,735 @@
...
+    if (!SUPPORTED_EXPORT_TYPES.has(exportType)) {
+        return { success: false, status: 400, message: "Unsupported export type." };
+    }
+    workflowIds = typeof workflowIds === 'string' ? JSON.parse(workflowIds) : workflowIds;
+    if (!Array.isArray(workflowIds)) workflowIds = [];
+    if (exportType === "grades" && !["json", "csv"].includes(normalizedGradeFormat)) {
+        return { success: false, status: 400, message: "Unsupported grade format. Use json or csv." };
+    }
+    if (userIds.length === 0) {
+        console.warn("Export aborted: No valid users selected.");
```
dennis-zyska, line 39: we should not use console.warn, we should use our logging mechanismn

backend/utils/helper/export.js
dennis-zyska, line 1: Can we split this file, it still feels too much in one file

dennis-zyska (changes requested):

backend/utils/helper/export.js
melolw: changed all console.warn

backend/utils/helper/export.js
melolw: split it into this and exportGrades.js which had the most functions. If it gets too long in future we can go ahead and do helpers for each export I think?

backend/utils/helper/export.js
melolw: i changed it, it should show the headers with empty rows now

backend/webserver/routes/export.js
melolw: if im not mistaken this is actually not checked for either of the exports. so i added it in loadExportRequestContext to either admin, or project owner, or yourself (so allowing to download own data)

frontend/src/components/dashboard/projects/export/StepSelectUsers.vue
melolw: i think i already changed this, after i asked about the scope and you told me all users of the whole platform should be included

frontend/src/components/dashboard/projects/export/StepConfirmDownload.vue
melolw: not really, i tried it in the user table with an "entries" column and the getStat call was getting too heavy to display
is there another way maybe?

frontend/src/components/dashboard/projects/ExportModal.vue
```diff
@@ -50,30 +53,57 @@
...
+        />
+        <!-- We get the desired document types as well as if non consenting users' edits should be included  -->
+      </div>
+      <div v-else-if="['studies'].includes(dataSelection.exportType)">
+        <StepOptionsStudies
+          :project-id="dataSelection.projectId"
+          v-model:selectedWorkflowIds="selectedWorkflowIds"
+          v-model:includeEmptyStudies="includeEmptyStudies"
+          v-model:includeDocumentFiles="includeStudyDocumentFiles"
+          v-model:includeGrades="includeStudyGrades"
```
dennis-zyska: The parent binds `v-model:includeGrades`, but this child declares the prop and emit as `includeScores`, so the "Include scores" switch never reaches `ExportModal`. `includeStudyGrades` stays `true`, so `includeGrades: true` is always posted at line 505 and scores are exported even with the switch off. Change the binding here to `v-model:includeScores="includeStudyGrades"` and leave the POST key at line 505 as `includeGrades`, which is what the backend reads.

backend/webserver/routes/export.js
```diff
@@ -775,22 +389,558 @@ module.exports = function (server) {
...
+                // not on the root document. Collect all copy IDs and query across them.
+                const copies = await server.db.models.document.findAll({
+                    where: { parentDocumentId: doc.id },
+                    attributes: ['id'],
+                    raw: true,
+                });
+                const allDocIds = [doc.id, ...copies.map(c => c.id)];
+
+                let [annotations, comments] = await Promise.all([
+                    server.db.models.annotation.findAll({ where: { documentId: allDocIds }, raw: true }),
```
dennis-zyska: These two queries are missing `deleted: false`, so soft-deleted annotations and comments end up in the export. The `comment_vote` query 12 lines below and the annotator-step queries in `processStudyBasedExport` both filter it, so this path is the odd one out. Add `deleted: false` here and to the `copies` lookup on line 423, which pulls deleted session copies for the same reason.

backend/webserver/routes/export.js
```diff
@@ -775,22 +389,558 @@ module.exports = function (server) {
...
+                                where: { documentId: step.documentId, studySessionId: session.id, studyStepId: step.id, deleted: false },
+                                raw: true,
+                            });
+
+                            if (shouldExcludeNonConsentingAnnotations) {
+                                const allUserIds = [...new Set([
+                                    ...annotations.map(a => a.userId),
+                                    ...comments.map(c => c.userId)
+                                ].filter(Boolean))];
+                                const consentedUsers = await server.db.models.user.findAll({
```
dennis-zyska: This is `getConsentedUserIds` written out by hand. The helper is already imported in this file and used in `processDocumentForExport`, so replace this block with `const consentedIds = await getConsentedUserIds(server, allUserIds);`. The same block is repeated again at line 741 for the editor steps.

backend/utils/helper/export.js
```diff
@@ -0,0 +1,418 @@
...
+ */
+async function loadExportRequestContext(server, { parsedProjectId, exportType, normalizedGradeFormat, userIds, workflowIds, currentUserId }) {
+    if (!Number.isInteger(parsedProjectId)) {
+        return { success: false, status: 400, message: "Missing projectId." };
+    }
+    if (!SUPPORTED_EXPORT_TYPES.has(exportType)) {
+        return { success: false, status: 400, message: "Unsupported export type." };
+    }
+    workflowIds = typeof workflowIds === 'string' ? JSON.parse(workflowIds) : workflowIds;
+    if (!Array.isArray(workflowIds)) workflowIds = [];
```
dennis-zyska, line 38: An empty `userIds` is rejected with a 400 four lines below, but an empty `workflowIds` falls through. For a `studies` export that reaches `processStudyBasedExport` as `workflowId: []`, which Sequelize turns into `IN ()`, so the user gets a 200 and an empty zip with no explanation. Reject it here the same way when `exportType === "studies"`.

backend/webserver/routes/export.js
```diff
@@ -775,22 +389,558 @@ module.exports = function (server) {
...
+     * filters by owner data sharing consent, and exports each document to the archive.
+     * @param {Object} server - The server instance providing database models.
+     * @param {number|string} projectId - The ID of the project to export.
+     * @param {string} baseFolderName - The root folder name inside the ZIP archive.
+     * @param {Object} archive - The archiver instance to append files to.
+     * @param {Array<number>} userIds - List of user IDs to filter documents by.
+     * @param {Array<number>} documentTypes - List of document types to include (0=PDF, 1=HTML, 2=Modal, 4=ZIP).
+     * @returns {Promise<void>}
+     */
+    async function processDocumentBasedExport(server, projectId, userIds, documentTypes, shouldExcludeNonConsentingEdits, shouldExcludeNonConsentingAnnotations, baseFolderName, archive) {
```
dennis-zyska: The split into `exportGrades.js` helped, but this route file is now 945 lines, up from 796 on dev, because the four new processors landed here rather than in a helper. `processDocumentForExport`, `processDocumentBasedExport`, `processStudyBasedExport` and `processUserBehaviourExport` are pure functions over `server` and `archive`, so they can move to `utils/helper/` next to the others and leave the route as the request handler plus a switch.

frontend/src/components/dashboard/projects/export/StepOptionsStudies.vue
```diff
@@ -0,0 +1,242 @@
...
+    <BasicForm
+      v-model="optionsData"
+      :fields="fields"
+    />
+  </div>
+</template>
+
+<script>
+import BasicForm from "@/basic/Form.vue";
+import BasicButton from "@/basic/Button.vue";
```
dennis-zyska, line 64: `BasicButton` is imported and registered but never used in the template.

backend/utils/helper/exportGrades.js
```diff
@@ -0,0 +1,333 @@
...
+ * processStudyBasedExport (grouped by session for a per-session scores.json).
+ *
+ * @param {Object} server - The server instance providing database models and Sequelize operators.
+ * @param {number} projectId - The project whose grades should be resolved.
+ * @param {Array<number|string>} userIds - The selected document owners.
+ * @param {Array<Object>} users - Full user records for the selected users.
+ * @param {boolean} shouldGenerateAliases - Whether student names should be anonymized.
+ * @param {boolean} hasPrivateInfoRight - Whether the requester may export real names.
+ * @param {Object} userMapping - Map of user IDs to generated aliases.
+ * @returns {Promise<{records: Array<Object>, criteriaReferencesByConfigId: Map<number, Object>}>}
```
dennis-zyska, line 217: There are two docblocks stacked here and only the second one attaches to the function, so the `@param` and `@returns` tags in the first are dropped by JSDoc and by IDE tooltips. Merge them into one block: keep the `options.sessionIds` explanation from the second and the tag list from the first.

backend/webserver/routes/export.js
```diff
@@ -775,22 +389,558 @@ module.exports = function (server) {
...
+        const sorted = [];
+        let current = items.find(item => item[prevKey] === null);
+        while (current) {
+            sorted.push(current);
+            current = items.find(item => item[prevKey] === current.id);
+        }
+        return sorted;
+    }
+
+    async function processStudyBasedExport(server, projectId, userIds, users, shouldGenerateAliases, hasPrivateInfoRight, userMapping, workflowIds, shouldIncludeEmptyStudies, shouldExcludeNonConsentingEdits, shouldExcludeNonConsentingAnnotations, shouldIncludeDocumentFiles, shouldIncludeGrades, shouldIncludeAiScores, baseFolderName, archive) {
```
dennis-zyska: This function has 16 positional parameters and no JSDoc, while every other new function in this file has one. Six of them are booleans in a row (`shouldIncludeEmptyStudies` through `shouldIncludeAiScores`), so a call site is easy to get wrong and hard to read. Group the flags into one `options` object and add a docblock describing it.

backend/utils/helper/export.js
```diff
@@ -0,0 +1,418 @@
...
+    try {
+        const parsed = typeof rawUserIds === 'string' ? JSON.parse(rawUserIds) : rawUserIds;
+        return Array.isArray(parsed) ? parsed : [];
+    } catch (e) {
+        server.logger.warn("Could not parse userIds:", rawUserIds);
+        return [];
+    }
+}
+
+async function attachTagNames(server, annotations) {
```
dennis-zyska, line 339: `attachTagNames`, `resolveIsAdmin` (line 297) and `createCsvRowsStream` (line 359) have no JSDoc, while the other functions in this file all do. `createJsonArrayStream` has a partial one that documents `fetchPage` but not `mapRow`, `pageSize`, or the return value. Worth a pass over the four before merge.

backend/webserver/routes/export.js
```diff
@@ -775,22 +389,558 @@ module.exports = function (server) {
...
+        }
+
+        for (const doc of filteredDocs) {
+            const docFolder = `${baseFolderName}/${doc.hash}`;
+            const docUserRoles = rolesMap[doc.userId] || [];
+            await processDocumentForExport(server, doc, docFolder, shouldExcludeNonConsentingEdits, shouldExcludeNonConsentingAnnotations, docUserRoles, archive);
+        }
+    }
+
+    function sortSteps(items, prevKey) {
```
dennis-zyska: This is the same walk as `workflow_step.getSortedWorkflowSteps` (`backend/db/models/workflow_step.js:61`), generalised over the key name. CARE keeps that one as a static method on the model, so the matching home for this is `study_step.getSortedStudySteps(studyId)`, which also makes it reusable outside the export.

dennis-zyska (changes requested):

PR #324, opened 2026-08-04
Feat 323 assignment bugs

karimouf:
## Overview

Three bug fixes around assignment revisions and assignment copying:
- The "max revisions" slider's unlimited sentinel collided with the newly-allowed "0 revisions" option
- The max-revisions error message exposed a confusing internal count
- Copying an assignment didn't record which assignment it was copied from

---

## Improvements
- Clarified the Maximum Revisions slider help text: N revisions means the original submission plus N resubmissions (e.g. 1 revision = 2 submissions total)
- Simplified the "maximum revisions reached" error message shown to users, dropping the internal (chainDepth/maxRevisions) count that could read as inconsistent with the revision terminology
- Assignment cards now show a "submissions left" badge and a due-date badge (moved into the card header, next to the upload button) instead of no indication at all
- Due-date badge turns orange within 24 hours of closing and red (with an "Overdue by ..." message) once the due date has passed, instead of always showing a plain, non-directional time difference
- "Submissions left" badge turns green once a user has 0 submissions remaining
- Users with the upload-for-others right see the due date for everyone but no longer see a personal "submissions left" count, since it isn't meaningful for them

---

## Bug fixes
- **Fixed:** the Maximum Revisions slider was fixed; the number described for each revision adding to your first submission we now can set the maximum revisions to zero, which means you only can upload one original submission
- **Fixed:** unlimited revisions is now set to -1 instead of a specific number
- **Fixed:** copying an assignment not recording its origin — `parentAssignmentId` is now set to the source assignment's ID when using "Copy Assignment", so copies stay traceable to their origin


backend/webserver/sockets/document.js
```diff
@@ -896,8 +896,8 @@ class DocumentSocket extends Socket {
                         transaction,
                     });
 
-                    // Check revision limit (0 = unlimited)
-                    if (assignment && assignment.maxRevisions > 0 && assignmentSubmissions.length >= assignment.maxRevisions) {
+                    // Check revision limit (-1 = unlimited)
+                    if (assignment && assignment.maxRevisions !== -1 && assignmentSubmissions.length >= assignment.maxRevisions) {
```
dennis-zyska: This check has an off-by-one bug; walk through it with `maxRevisions = 1` (meaning: 1 revision allowed, so 2 submissions total).

A user has already submitted once. `assignmentSubmissions.length` is `1`. The condition is `1 >= 1`, **true**, it throws "Revision limit reached," blocking their 2nd submission, even though 2 submissions should be allowed. `maxRevisions` counts *revisions after the original*, not total submissions.I think it needs `maxRevisions + 1`. same in line ~1168 (`chainDepth >= assignment.maxRevisions`)

dennis-zyska (changes requested):

backend/db/models/assignment.js
```diff
@@ -32,16 +32,16 @@ module.exports = (sequelize, DataTypes) => {
...
 				class: "custom-slider-class",
-				min: 1,
+				min: 0,
 				max: 20,
 				step: 1,
 				unit: "revision(s)",
 				unlimitedAtMax: true,
 				unlimitedLabel: "unlimited",
-				unlimitedStoredValue: 0,
+				unlimitedStoredValue: -1,
```
dennis-zyska, line 41: The unlimited value for `maxRevisions` changed from `0` to `-1` here, but `AssignmentSubmissionsModal.vue` (line 154) still checks `maxRevisions === 0` for unlimited and compares depth without the `+1` offset used elsewhere. That means a zero-revision assignment now looks unlimited there, while an unlimited assignment gets blocked right away. Update `isRevisionLimitReached` in that file to check `-1` for unlimited and compare against `maxRevisions + 1`, matching `Submissions.vue`.

backend/webserver/sockets/document.js
```diff
@@ -1210,7 +1210,7 @@ class DocumentSocket extends Socket {
 
                     if (chainDepth >= assignment.maxRevisions) {
```
dennis-zyska, line 1211: This has an off-by-one bug for the check above at line 901: `chainDepth >= assignment.maxRevisions` should be `chainDepth >= assignment.maxRevisions + 1`. Right now a user with `maxRevisions = 1` gets blocked on their 2nd submission instead of their 3rd.

frontend/src/components/dashboard/Submissions.vue
```diff
@@ -97,8 +117,14 @@ import { getTimeDiffString } from "@/assets/utils";
 export default {
   name: "DashboardSubmission",
   subscribeTable: ["assignment", "submission", "user", "document"],
-  components: { Card, BasicButton, AssignmentUploadModal, AssignmentSubmissionsTable, ConfirmModal },
+  components: { Card, BasicButton, LoadIcon, AssignmentUploadModal, AssignmentSubmissionsTable, ConfirmModal },
   computed: {
+    isAdmin() {
+      return this.$store.getters["auth/isAdmin"];
```
dennis-zyska, line 123: isAdmin is added here but nothing in this file calls this.isAdmin

frontend/src/components/dashboard/Submissions.vue
```diff
@@ -186,10 +212,42 @@ export default {
...
+      const msLeft = new Date(assignment.end) - new Date();
+      return msLeft > 0 && msLeft <= 24 * 60 * 60 * 1000;
+    },
+    timeBadgeClass(assignment) {
+      if (this.isAssignmentOverdue(assignment)) return "bg-danger text-white";
+      if (this.isAssignmentClosingSoon(assignment)) return "bg-warning text-dark";
+      return "bg-light text-dark border";
+    },
+    timeBadgeText(assignment) {
+      console.log(assignment)
```
dennis-zyska, line 246: Leftover console.log(assignment) on this line should be removed before merging.

dennis-zyska (changes requested):

dennis-zyska (approved):

PR #333, opened 2026-08-09
fix : restrict public templates list and view to importable types

### LINKED ISSUE #307: [BUG] public templates modal should only list importable templates
# Describe the Bug:  

Public Templates lists all public templates and allows opening any of them read-only. Non-admins should only see templates they are allowed to copy (document types), not email templates they cannot import.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

1. Log in as non-admin user.
2. Templates → Public Templates.
3. Observe email templates listed; use View on a template you cannot copy.

# Expected Behavior:  

Only show templates the user can import. No read-only access to templates they cannot own/copy.

# Actual Behavior:  

All public templates shown. View opens `/template/:id` even when copy is denied server-side.

mohammadsherif0:
## Description
Fixes #307: non-admins could see and open public email templates in Public Templates even though they cannot import them. Listing and content view now follow the same importable-type rules as copy (types 4/5 for non-admins).

## Improvements
Public Templates modal shows only importable public templates for non-admins (types 4 and 5); admins still see all public templates.
`templateGetContent` blocks non-owners from viewing public email templates unless they are admin.


backend/webserver/sockets/template.js
```diff
@@ -92,10 +92,15 @@ class TemplateSocket extends Socket {
     
     const isOwner = template.userId === this.userId;
     const isPublicFromOthers = template.public === true && !isOwner;
+    const isAdmin = await this.isAdmin();
+    const isEmailType = [1, 2, 3, 6, 7].includes(template.type);
```
bingobongomann, line 96: I think this is the 9th or 10th time that we have the [1,...,7], [1,2,3,6,7] and [4,5] arrays in the template code. I think we should move this into some static lists like emailTemplateTypes = [1,2,3,6,7], otherTemplateTypes = [4,5], allTemplateTypes = emailTemplateTypes + otherTemplateTypes 
And then use those statics lists. otherwise we have to adjust many different locations when adding new types in the future

backend/webserver/sockets/template.js
bingobongomann, line 74: The docstring should reflect the rejection behavior for non-owners non-admins on email templates

bingobongomann (changes requested):
Two small improvements. One not even specific to this PR but generally to the template types arrays. 

I could not reproduce the issue on the current dev branch

bingobongomann (approved):

backend/db/models/template.js
```diff
@@ -510,4 +517,8 @@ module.exports = (sequelize, DataTypes) => {
         }
     );
     return Template;
-}
\ No newline at end of file
+}
+
+module.exports.emailTemplateTypes = emailTemplateTypes;
```
dennis-zyska: These three lines attach the arrays to the exported factory function, which nothing reads. The model loader calls the factory and registers the returned class, so every consumer goes through `models["template"].emailTemplateTypes` (the static on line 16). Please delete lines 522-524.

docs/source/for_developers/frontend/components/templates.rst
```diff
@@ -147,7 +147,8 @@ Here is a concrete example for adding a new placeholder (e.g. ``studyEndDate`` f
 
    - If the new placeholder is tied to a new template type, also update:
 
-     - The template type dropdown in ``frontend/src/components/dashboard/templates/TemplateModal.vue``.
+     - ``emailTemplateTypes`` / ``otherTemplateTypes`` in ``backend/db/models/template.js``.
+     - The template type dropdown options in ``backend/db/models/template.js`` ``fields``.
```
dennis-zyska: The deleted bullet is still needed: `TemplateModal.vue:60` hardcodes `[4, 5]` to filter the type options for non-admins, so a new type has to be added there too. Please keep the `TemplateModal.vue` bullet alongside the two new ones.

frontend/src/components/dashboard/templates/PublicTemplatesModal.vue
```diff
@@ -70,8 +70,11 @@ export default {
         .filter(t => t.userId === this.userId && !t.deleted);
     },
     publicTemplates() {
+      const importableTypes = [4, 5];
```
dennis-zyska, line 71: This adds a sixth hardcoded `[4, 5]` in the frontend, in the same PR that centralized these arrays on the backend. Please define one shared frontend constant for the importable types and use it here, so a new template type only needs one frontend change.

dennis-zyska (changes requested):

docs/source/for_developers/frontend/components/templates.rst
mohammadsherif0: The TemplateModal filter does belong on the new-type checklist, but that whole section mixed “add a placeholder” and “add a type” with no real split. I’ll rewrite the section so those are two separate lists, and keep the `TemplateModal.vue` non-admin filter on the type one.

frontend/src/components/dashboard/templates/PublicTemplatesModal.vue
mohammadsherif0: Done in #334

mohammadsherif0:
This branch is merged into #334 and the conflicts with dev have been resolved there. Should I close this pull request?

dennis-zyska:
Yes, we can merge it through the other branch, but it has some merge conflicts, please have a look

PR #334, opened 2026-08-09
Feat 308 email template admin access

### LINKED ISSUE #308: [BUG] template access persists in UI after admin role is removed
# Describe the Bug:  

Users who previously had admin and created or imported email/document templates still see and edit those templates after admin is removed, including updates from source templates they should no longer access.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

1. User with admin creates/imports email template.
2. Remove admin role; keep user role only.
3. Do not refresh and open Templates / template editor.
4. User can still view/edit email templates and may receive source updates.

# Expected Behavior:  

After role change, user only sees templates allowed by `getUserFilter` (doc types 4/5 for non-admin). Email templates hidden. Source-template broadcasts stop for unauthorized users.

# Actual Behavior:  

Stale Vuex `table/template` data allows continued access.

mohammadsherif0:
## Description

Fixes #308: after admin is removed, a user who still owns email templates (types 1/2/3/6/7) could view/edit/save them without a hard refresh. Email template actions now re-check admin, and the rights cache refreshes when `rolesUpdatedAt` changes.

Please review **#333 first**: it already restricts public list/view to importable types (4/5) for non-admins; this PR continues that policy for owned email templates and for mid-session demotion.

## Improvements
- Email `getContent` / `editContent` / `closeTemplate` / `updateFromSource` deny non-admins (owners included).
- `expandBroadcastFilter` only tracks sources for importable copies (types 4/5), so non-admins are not subscribed to email sources.
- `getUserRolesById` ignores soft-deleted role rows.
- `User.cache.clear()` after role updates so `findByPk` sees the new `rolesUpdatedAt`.
- `isAdmin` / `hasAccess` compare DB `rolesUpdatedAt` (ms) to the cached stamp and reload when it changes.
- Cache both allowed and denied rights (`!== undefined`), so a deny is not looked up again every time.

## Idea / open question
Can all Socket subclasses on one connection share a single rights cache? Right now each keeps its own `userInfo`.


backend/webserver/sockets/template.js
```diff
@@ -92,10 +92,15 @@ class TemplateSocket extends Socket {
     
     const isOwner = template.userId === this.userId;
     const isPublicFromOthers = template.public === true && !isOwner;
+    const isAdmin = await this.isAdmin();
+    const isEmailType = [1, 2, 3, 6, 7].includes(template.type);
```
dennis-zyska, line 96: This PR adds four more copies of the `[1, 2, 3, 6, 7]` email-type array, which already appears seven times in this file. Consider extracting an `EMAIL_TEMPLATE_TYPES` constant so the list can't drift between call sites.

backend/webserver/sockets/template.js
```diff
@@ -92,10 +92,15 @@ class TemplateSocket extends Socket {
...
     
     if (!isOwner && !isPublicFromOthers) {
       throw new Error("You can only view templates that you own or public templates from others");
     }
+    if (!isAdmin && isEmailType) {
```
dennis-zyska, line 101: These new checks only guard the bespoke socket handlers. Template.js's own `beforeUpdate` hook (line 489) checks `options.callerUserId`, but `updateData` in app.js only ever sets `options.context.currentUserId` (grep confirms `callerUserId` is never set anywhere), so the hook's ownership/admin check never runs. Any user can bypass every gate in this PR by calling `appDataUpdate` directly on the `template` table with an email template's id.

backend/webserver/sockets/template.js
```diff
@@ -194,6 +199,10 @@ class TemplateSocket extends Socket {
       throw new Error("Copied templates cannot be edited");
     }
 
+    if ([1, 2, 3, 6, 7].includes(template.type) && !(await this.isAdmin())) {
```
dennis-zyska, line 202: `addContent` (templateAddLanguageContent, same file) still creates a new `template_content` row for any template the caller owns, with no `isAdmin()`/email-type check. A demoted user who still owns an email template can use it to write fresh content, bypassing this exact restriction.

backend/webserver/sockets/template.js
```diff
@@ -589,6 +598,10 @@ class TemplateSocket extends Socket {
     const template = await this.models["template"].getById(data.templateId);
     if (!template) return;
 
+    if ([1, 2, 3, 6, 7].includes(template.type) && !(await this.isAdmin())) {
```
dennis-zyska, line 601: `getAllPlaceholders` and `getLanguages` still return template-type and placeholder metadata for templates the caller merely owns, without the admin+email-type gate used here. They should get the same check so a demoted owner can't read email-template structure.

backend/db/models/user.js
```diff
@@ -475,6 +475,14 @@ module.exports = (sequelize, DataTypes) => {
                     }
                 }
             }
+
+            // SequelizeSimpleCache keeps User rows indefinitely (ttl: false). After roles
+            // change, clear it so the next findByPk sees the new rolesUpdatedAt.
+            options.transaction.afterCommit(() => {
```
dennis-zyska, line 481: User.cache has ttl:false (eternal caching), so only an explicit clear() invalidates it, and this is the only call site that does one. The user_role_matching afterCreate/afterDestroy hooks (user_role_matching.js:112-134) also update rolesUpdatedAt whenever roles change via assignUserRoles (the import path, user.js around line 639/645), but never clear the cache there, so a demoted admin imported via that path would still pass isAdmin() on a stale cached row. Move the cache.clear() into those hooks instead so every mutation path is covered.

dennis-zyska (changes requested):

Issue #212, opened 2026-05-19
[USER STORY] Dark mode for CARE


akash9676:
## Description 

As a CARE user who reads and works on documents for long periods, user want to turn on a dark screen theme so that the app is easier on eyes.

## Acceptance Criteria 

- Users can switch between light and dark mode from the app.
- The chosen mode is remembered the next time they open CARE.
- Dark mode applies across the main areas of the app (home/dashboard, reading and annotating, writing, and settings).
- Text, buttons, and menus stay easy to read in both modes.
- Turning dark mode on or off does not change how the app works only how it looks.

## Estimated effort (Story Points) 

5

## Tasks 

- [ ] Add a light/dark mode toggle where users can find it easily (e.g. settings or profile)
- [ ] Save the user’s choice so it persists between sessions
- [ ] Apply dark styling to the main layout (navigation, sidebars, content areas)
- [ ] Apply dark styling to the reading and annotating views
- [ ] Apply dark styling to the writing/editor views
- [ ] Apply dark styling to dashboard and settings screens
- [ ] Check readability of text, buttons, forms, and messages in both modes
- [ ] Test that existing features still work in both light and dark mode

## Source Branch 

dev

PR #336, opened 2026-08-09
Feat 212 dark mode

theultimatewasteofspace:
## Description

Implements a dark theme for CARE (Issue #212). The theme is applied by setting
`data-bs-theme` on `<html>`, so Bootstrap 5.3's own theming handles most of the
application; the rest of this PR covers the surfaces Bootstrap does not reach.
Users switch theme from the avatar dropdown in the topbar, and the choice is
saved as the `app.theme.mode` setting.

The bulk of the diff is migrating hardcoded colour values across basic and
feature components to Bootstrap theme variables so they respond to the theme
instead of staying fixed.

## New User Features

- **Theme toggle** in the avatar dropdown, labelled for the theme it switches to.
- **Theme resolution on first load** — the saved user setting takes precedence,
  then a locally cached value, then the browser's `prefers-color-scheme`. This
  means logged-out pages and first paint already show the correct theme.
- **Per-document PDF theme toggle** in the PDF toolbar, independent of the app
  theme, so an individual document can be read in light without leaving dark
  mode. The choice persists across reloads and resets when the app theme changes.
- **Dark PDF rendering** — page canvases are inverted and tinted cool so they sit
  with the rest of the dark UI rather than reading warm. Annotation highlights
  switch to `screen` blending so they stay visible over an inverted page, where
  `multiply` would render them nearly invisible.
- **Configurable dark accent colour** — new `theme.dark.accentColor` setting
  (type `color`, Interface → Branding), defaulting to CARE's logo yellow
  `#e3d5a8`. Hover, active and switch-knob colours are derived from it
  automatically.

## New Dev Features

- Three exported helpers in `frontend/src/assets/utils.js`:
  - `applyTheme(mode)` — sets `data-bs-theme` and caches the choice locally.
  - `resolveTheme(saved)` — implements the setting → cache → OS precedence.
  - `shadeColor(hexColor, amount)` — shifts a colour towards black or white,
    whichever is visible against it, for deriving hover and active states.
- A `--care-accent` / `--care-on-accent` / `--care-accent-hover` /
  `--care-accent-active` variable family, recomputed when settings change. New
  components should use these rather than referencing the accent colour directly.
- `color-scheme` is tied to the selected theme, so browser-native controls and
  scrollbars follow CARE rather than the operating system.

## Improvements

- Basic and feature components migrated from hardcoded hex values to Bootstrap
  theme variables.
- Quill editor themed: toolbar, container, editor body, placeholder text, picker
  dropdowns, and toolbar icon stroke/fill.
- Annotator icons switched to outline variants for legibility on dark.
- Custom scrollbars themed.
- Disabled buttons and nav links given higher opacity, since they lose more
  contrast against dark than against white.
- Printing never inverts the PDF canvas — a dark page wastes ink and reads badly.
- White PDF pages are slightly dimmed inside a dark app, cutting glare without
  distorting figure colours.

## Bug Fixes

- `PDFToolbar.vue`: the toolbar toggle's tooltip was written as a literal string
  rather than a binding, so the raw expression was shown as tooltip text.
- `PDFPage.vue`: the page separator declared `border-bottom-style` with no width
  or colour, leaving the rendered border to browser defaults.
- `Topbar.vue`: the project selector's `<li>` rendered even when the button
  inside it was hidden, leaving stray spacing in the navbar.
- `App.vue`: switch-knob generation ran outside the guard for a missing accent
  setting, which would throw before the setting had loaded.
- `Comment.vue`: a malformed colour declaration made the whole rule invalid, so
  comment text lost its colour in both themes.
- Backgrounds left hardcoded in otherwise-themed rules, which produced
  light-on-light text in dark mode: the assessment feedback box and score
  dropdown, the floating info panel, the manage-studies selected-items list and
  confirmation panel, and the workflow graph tooltip.
- Two rules used a text-colour variable as a background, giving poor contrast in
  both themes: the information modal's outline-secondary hover state and the
  placeholder type badge.

## Known Limitations

- The configurable accent currently applies to a subset of dark-mode controls —
  primary buttons, pagination and checkboxes. Form focus rings, selected cards,
  stepper indicators and icon hover states still resolve Bootstrap's default
  blue, so a dark-mode form can show a yellow button beside a blue focus ring.
  Deciding how far the accent should extend is worth settling before the next
  branding change.
- The accent is configurable for dark mode only; light mode uses Bootstrap's
  default blue.
- Dark PDF pages are produced by a CSS filter inversion rather than a re-render,
  so figure and image colours shift. The per-document light toggle is the escape
  hatch where colour accuracy matters.
- Quill, PDF.js and bootstrap-vue-3 do not follow `data-bs-theme`. Their surfaces
  are covered by explicit overrides in `global.css`, so new UI built on those
  libraries will not inherit the theme automatically.
- The theme cache is per browser, so a user on a second browser sees the OS
  preference until their settings load.

## Future Steps

- Settle the accent scope above and apply it consistently.
- Extend accent configuration to light mode so branding is consistent across both
  themes.
- Consider following OS theme changes live, rather than reading the OS preference
  only when no setting or cache exists.
  
  ## NOTE (added)

Note for reviewers: run make db after checking out, or the accent setting won't exist and buttons will fall back to the default.


frontend/src/basic/modal/skills/InputFiles.vue
```diff
@@ -305,11 +305,11 @@ export default {
 
 <style scoped>
 .selection-summary-custom {
-  background-color: #e7f3ff;
+  background-color: var(--bs-primary-bg-subtle, #e7f3ff);
   border: 1px solid #007bff;
```
dennis-zyska: `color` was migrated to `var(--bs-primary, #007bff)` but `border` on the line above keeps the raw `#007bff`, so in dark mode the border stays Bootstrap blue while the text switches to the accent yellow, clashing on the same box. Migrate the border to `var(--bs-primary, #007bff)` too.

frontend/src/basic/navigation/Topbar.vue
```diff
@@ -187,6 +192,10 @@ export default {
     consentEnabled() {
       return this.$store.getters["settings/getValue"]("app.config.consent.enabled") === "true";
     },
+    isDarkMode() {
```
dennis-zyska, line 195: isDarkMode only checks the saved setting and localStorage, unlike App.vue's initTheme() which also falls back to prefers-color-scheme. On a first visit with no saved setting and an OS dark preference, the page renders dark but this computed still returns false, so the toggle mislabels itself and the first click is a no-op. Mirror initTheme()'s three-tier check here too.

frontend/src/components/annotator/pdfViewer/Highlights.vue
```diff
@@ -220,22 +220,22 @@ export default {
         if (tag) {
           switch (tag.colorCode) {
             case "success":
-              return "009933";
+              return "4dd47a";
```
dennis-zyska: getColor() returns entirely new hex values (009933->4dd47a, e05f5f->f08a8a, etc.) instead of a var()-based substitution, and the function has no theme check, so every annotation highlight color changes in light mode too.

dennis-zyska (changes requested):
Also check the linter comments

frontend/src/basic/modal/skills/InputFiles.vue
theultimatewasteofspace: You're right. It turned out to be systematic rather than one line, the original sweep matched `border-color:` but not the `border:` shorthand, so around 40 more were hiding.

Migrated the structural greys to `var(--bs-border-color, ...)`. Left the semantic ones (danger red, warning yellow, the annotation glow) and the decorative `border-left` accent bars, since converting those to `--bs-primary` would turn them yellow in dark mode, which isn't what they mean.

frontend/src/basic/navigation/Topbar.vue
theultimatewasteofspace: I've mirrored the three-tier check, but pulled the resolution into a `resolveTheme` helper in utils.js rather than repeating it, so `initTheme` and `isDarkMode` both call the same thing and can't drift apart again. I hope that's okay but happy to inline it in both places instead if you'd prefer that.

frontend/src/components/annotator/pdfViewer/Highlights.vue
theultimatewasteofspace: Reverted them entirely. I'd changed the colours while chasing why highlights were invisible on dark pages, before finding the actual cause which was the highlight layer uses mix-blend-mode: multiply, which works like a marker pen on white paper but does almost nothing over dark. With that switched to screen in dark mode, the original colours read fine now so the colour change wasn't needed at all

pdji1602003:
sc

In dark mode, the text on the button is hard to see when the mouse hovers on the button.

frontend/src/components/annotator/pdfViewer/PDFViewer.vue
```diff
@@ -113,9 +115,15 @@ export default {
       zoomFormData: {
         zoom: 1.0,
       },
+      pdfDarkMode: localStorage.getItem("care.pdfDark") !== null
```
pdji1602003, line 119: Here we only use one localStorage key, care.pdfDark and this would make one document’s PDF theme choice affect later PDF. I am not sure if this is a desirable effect. If we want it to be per document only, maybe we could try `care.pdfDark.${documentId}.`

frontend/src/App.vue
```diff
@@ -191,6 +192,26 @@ export default {
     }
   },
   watch: {
+   "$store.state.settings": {
```
pdji1602003: I think the following four CSS variables should be defined in the global.css as well (html[data-bs-theme="dark"]) as now we only assign the values when the setting is loading, so on first paint, these values would be undefined.

pdji1602003 (dismissed):

pdji1602003:
The accent colour of the menu buttons does not get changed to yellow when the button is clicked.
sc

frontend/src/components/annotator/pdfViewer/PDFViewer.vue
theultimatewasteofspace: Good point, I did think about this and decided to lean towards keeping it global because it's more of a reading preference than a property of the document, so if someone finds inverted PDFs hard to read, they probably want light for everything rather than retoggling on each file. Same reasoning as the app theme being one setting rather than per page. But I'll reconsider it and do some research on how other platforms do it and change accordingly :)

frontend/src/App.vue
theultimatewasteofspace: Done. I hit the same thing reproducing your other comment. All four now have defaults in the html[data-bs-theme="dark"] block, so they're set before the settings load rather than being undefined until then.

theultimatewasteofspace:
@pdji1602003 thanks for these!

The hover one was the accent colour setting and it comes with a migration, so without running `make db` the CSS variables never get a value and the hover has nothing to fall back on. My fault for not mentioning that in the description. I've added defaults in global.css so it works either way now.

The blue dropdown item was a separate gap. Bootstrap uses its own variables for dropdown active states rather than primary, so it wasn't picking up the accent. Fixed, along with disabled buttons which had the same problem.

frontend/src/assets/styles/global.css
```diff
@@ -21,3 +21,209 @@
...
+}
+
+[data-bs-theme="dark"] .form-check-input:checked {
+  background-color: var(--care-accent);
+  border-color: var(--care-accent);
+}
+
+[data-bs-theme="dark"] .form-check-input:focus {
+  border-color: var(--care-accent);
+  box-shadow: 0 0 0 0.25rem rgba(227, 213, 168, 0.25);
```
dennis-zyska: The focus ring here hardcodes `rgba(227, 213, 168, 0.25)`, which is the default accent. Every other accent colour in this block reads `var(--care-accent)`, so if an admin changes `theme.dark.accentColor` the checkbox border turns the new colour but the ring around it stays yellow. Use `color-mix(in srgb, var(--care-accent) 25%, transparent)` instead.

frontend/src/components/annotator/Annotator.vue
```diff
@@ -318,7 +318,7 @@ export default {
           id: 'toggle-nlp',
           icon: 'robot',
           title: this.nlpActive ? 'Deactivate NLP support' : 'Activate NLP support',
-          color: this.nlpActive ? '#097969' : '#777777',
+          color: this.nlpActive ? '#12a085' : '#9aa3b0',
```
dennis-zyska: These two colours are swapped for new fixed values with no theme check, so the NLP button changes in light mode too. Every other icon colour in this PR moved to a variable instead, for example `var(--bs-danger-text-emphasis)` in `Study.vue:45` and `var(--bs-secondary-color)` in `annotator/sidebar/Sidebar.vue:40`. Do the same here: add a `--care-nlp-active` / `--care-nlp-idle` pair to `global.css` with the old values in `:root` and your new ones under `html[data-bs-theme="dark"]`, then read them from this line.

frontend/src/components/annotator/pdfViewer/PDFViewer.vue
```diff
@@ -130,6 +138,10 @@ export default {
     },
   },
   watch: {
+    appTheme(newVal) {
+      this.pdfDarkMode = newVal === "dark";
+      localStorage.removeItem("care.pdfDark");
```
dennis-zyska, line 145: `getValue("app.theme.mode")` is `undefined` until the `appSettings` socket message arrives, so `appTheme` changes from `undefined` to its real value on every page load. That fires this watcher, which calls `localStorage.removeItem("care.pdfDark")` and wipes the saved choice before the reader has touched anything. Change the signature to `appTheme(newVal, oldVal)` and start the body with `if (oldVal === undefined) return;` so it only reacts to a real theme change.

frontend/src/App.vue
```diff
@@ -191,6 +192,26 @@ export default {
...
+      handler() {
+        const saved = this.$store.getters["settings/getValue"]("app.theme.mode");
```
dennis-zyska, line 216: `app.theme.mode` is read here and written from `Topbar.toggleTheme`, but no migration creates a `setting` row for it, unlike `app.locale` and `projects.default`. That means there is no instance default an admin can set, and the key never appears in the settings dashboard. Add it to the new migration with `allowUserOverride: true`, otherwise the `user_setting` hook will start rejecting every user toggle.

frontend/src/components/annotator/sidebar/card/Comment.vue
```diff
@@ -513,7 +513,7 @@ export default {
 
 <style scoped>
 .comment {
-  color: #666666;
+  color: var(--bs-secondary-color, #666666);
```
dennis-zyska, line 518: The malformed declaration in this rule is `box-shadow: 2px, 3px lightgrey;` two lines down, and it is still there. CSS drops only the bad declaration, not the whole rule, so `color` was applying fine before this change. Fix it to `box-shadow: 2px 3px lightgrey;` or drop the line, and correct the Bug Fixes entry that describes this as a lost colour.

frontend/src/components/dashboard/assignments/metadata/StepMapping.vue
```diff
@@ -172,7 +172,7 @@ export default {
   border: 1px dashed #b9b9b9;
   border-radius: 0.5rem;
   padding: 0.9rem;
-  background: #fafafa;
+  background: var(--bs-tertiary-bg, #fafafa);
```
dennis-zyska, line 175: The background in this rule moved to a variable but `border: 1px dashed #b9b9b9` three lines up did not, so the panel keeps a light grey dashed border on a dark background. The same gap is in `dashboard/users/ImportModal.vue:507` (`border: 2px dashed #ccc`), where the hover background was migrated but the border was not. Both are structural greys, so `var(--bs-border-color, ...)` fits.

dennis-zyska (changes requested):

frontend/src/assets/styles/global.css
theultimatewasteofspace: Good catch, fixed. I'd assumed the rgba() had to stay literal since a CSS variable can't go inside it, color-mix is the obvious answer and I hadn't thought of it. Thanks.

frontend/src/components/annotator/Annotator.vue
theultimatewasteofspace: Both fixed. The focus ring uses color-mix now so it follows the accent, and the NLP icon reads from a --care-nlp-active / --care-nlp-idle pair with the original values in :root.

frontend/src/components/annotator/pdfViewer/PDFViewer.vue
theultimatewasteofspace: Good catch, fixed. Added the oldVal guard so it only reacts to a real theme change rather than the initial settings load.

frontend/src/App.vue
theultimatewasteofspace: Added both to the migration: `app.theme.mode` with `allowUserOverride: true` and an instance default, `theme.dark.accentColor` with `false` since it's admin-only.

frontend/src/components/annotator/sidebar/card/Comment.vue
theultimatewasteofspace: You're right, the comma made that one declaration invalid, so only the shadow was dropped and the colour was applying fine. Removed both malformed shadows (Comment.vue and Adder.vue) rather than correcting the syntax, since nobody has seen the intended shadow and adding one now would be a visual change unrelated to theming. Corrected the Bug Fixes entry too.

Also spotted IconBoostrap[disabled] in Annotator.vue, the selector is misspelled, so that rule never matches. Left it alone since it's unrelated, but worth flagging.

frontend/src/components/dashboard/assignments/metadata/StepMapping.vue
theultimatewasteofspace: Fixed both, plus two more from re-running the sweep wider, a `border: 1px solid gray` in Table.vue that the hex-only grep had missed, and one in the socketprofiler modal that landed after my original pass.

frontend/src/components/annotator/sidebar/card/Comment.vue
theultimatewasteofspace: removed

dennis-zyska (approved):

Issue #171, opened 2026-04-20
[BUG] Wrong tooltip on toggle settings — all show "Activate/Deactivate NLP support"


akash9676:
# Describe the Bug:

In the settings dashboard, the tooltip shown when hovering the toggle for `system.mailService.smtp.auth.enabled` (System → Mail Service → SMTP) says "Activate/Deactivate NLP support", which has nothing to do with SMTP auth.

The same wrong tooltip appears on **every** toggle (boolean) setting in the dashboard, not just this one.

# Reproducibilty:

- [x] Yes
- [ ] No
- [ ] Sometimes

# Steps to Reproduce the Behavior:

1. Log in as an admin.
2. Open the settings dashboard.
3. Go to **System → Mail Service → SMTP** and hover over the toggle for `auth.enabled`.
4. Check the tooltip.
5. Try the same on other toggle settings (e.g. auth providers, 2FA, logo, etc.) — same tooltip everywhere.

# Expected Behavior:

The tooltip should match the setting being toggled (or just not mention NLP on non-NLP settings).

# Actual Behavior:

All toggle settings show the tooltip "Activate/Deactivate NLP support" on hover, regardless of what the setting actually does.

# Screenshots:

<img width="1127" height="561" alt="Image" src="https://github.com/user-attachments/assets/cc779571-3b76-474b-b347-a21319d4b32e" />

# System Information:

- **Device (OS/Type):** Windows 10
- **Browser (+ Version):** Google Chrome (latest)
- **CARE Software (Version/Branch):** `dev`

# Additional Information:

- the boolean/toggle branch in `frontend/src/components/dashboard/settings/SettingItem.vue` (the `<input role="switch">` for `setting.type === 'boolean'`).

junaidferoz:
These are the changes I made to eliminate this bug

1. First I removed the hardcoded tooltip message and replaced with this 

```
  class="form-check-input" 
                            role="switch" 
                            title="Activate/Deactivate NLP support"
                            :title="getBooleanToggleTooltip(setting)"
                            type="checkbox"
```

2. Then I made sure to get the tooltip message from the description

```
 getBooleanToggleTooltip(setting){
 const label = (setting.description && setting.description.trim()) || setting.key;
      return `${label}`;
```

Note: I know the function name is misleading, so i am waiting for the requested changes to be pushed into dev first, then do the necessary changes

PR #341, opened 2026-08-13
fix: add tooltip on hover for help icon in settings

romila25:
This pull request adds a tooltip which appears when you hover over a help icon in settings dashboard.
It resolves the issue: https://github.com/UKPLab/CARE/issues/171

### DIFF
--- a/frontend/src/basic/form/Help.vue
+++ b/frontend/src/basic/form/Help.vue
@@ -1,6 +1,7 @@
 <template>
   <span
     v-if="help"
+    ref="tooltip"
     v-tooltip="!!help"
     class="btn btn-sm mt-0 pt-0 border-0"
     :class="buttonClass"
--- a/frontend/src/basic/form/Help.vue
+++ b/frontend/src/basic/form/Help.vue
@@ -18,6 +18,7 @@
 
 <script>
 import LoadIcon from "@/basic/Icon.vue";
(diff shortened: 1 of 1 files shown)


frontend/src/basic/form/Help.vue
```diff
@@ -42,12 +43,23 @@ export default {
...
   },
-  mounted() {
-    //new Tooltip(this.$refs.tooltip, {trigger: "click"});
-  }
-}
+
+   mounted() {
+    if (!this.$refs.tooltip) return;
+
+    this.tooltipInstance = new Tooltip(this.$refs.tooltip, {
```
dennis-zyska: This manually instantiates and disposes a Bootstrap `Tooltip`, duplicating the existing `tooltip` directive in `@/assets/tooltip.js` that `Table.vue`, `table/Icon.vue`, and `table/Badge.vue` already use for the same `data-bs-toggle="tooltip"` markup present in this template. Prefer `v-tooltip="!!help"` with `import { tooltip } from "@/assets/tooltip.js"` and drop the manual `mounted`/`beforeUnmount` logic. We should use a single way to align the solutions and don't have too many duplicated codes.

dennis-zyska (changes requested):

dennis-zyska (approved):

PR #367, opened 2026-09-08
Fix 361 access filter split

### LINKED ISSUE #310: [BUG] appDataUpdate bypasses authorization for partial special keys updates
# Describe the Bug

`AppSocket.updateData()` immediately updates records containing `id` and one of these fields:

`deleted`, `closed`, `public`, `end`, `disable`, `enabled`.

This early return bypasses the authorization logic that follows it. A user may therefore update a record they do not own by submitting its ID with one of these fields.

# Reproducibilty

- [x] Yes
- [ ] No
- [ ] Sometimes

# Steps to Reproduce the Behavior

1. Sign in as a non-admin user.
2. Obtain the ID of a record owned by another user.
3. Emit `appDataUpdate` with that ID and a special field, for example:

```json
{
  "table": "document",
  "data": {
    "id": 123,
    "deleted": true
  }
}
```

4. Observe that `updateById()` is called before the ownership check.

# Expected Behavior

Every update is authorized before changing the record. Partial updates remain supported, and required-field validation does not incorrectly reject them.

Models `beforeUpdate ` hooks can be a possible way to handle this, which may contain conditional runs for those special keys updates only.

# Actual Behavior

Updates containing one of the special fields return early and bypass the authorization check.

# Screenshots

Not applicable.

# Additional Information

Relevant file: `backend/webserver/sockets/app.js`, `AppSocket.updateData()`.

Acceptance criteria:

- [ ] No update path bypasses authorization.
- [ ] Partial updates such as delete, publish, close, disable, and enable continue to work.
- [ ] Required fields are enforced when creating records without requiring complete payloads for partial updates.
- [ ] Model-specific authorization is enforced before mutation.

### LINKED ISSUE #361: [BUG] updateData performs no ownership check on row updates
# Describe the Bug:  

updateData in backend/webserver/sockets/app.js updates a row by id without checking that the caller owns it. Any authenticated user can modify rows they don't own across every model exposed via appDataUpdate. The existing checkUserAccess guard only fires when the payload sets a userId and protects against reassigning ownership, not against editing a row owned by someone else.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

1. Log in as a low-privilege user (e.g. guest).
2. Identify the id of a record owned by another user (e.g. an admin-owned, non-public document).
3. Send an appDataUpdate for that record's id changing a state field (e.g. deleted, public, closed).
4. The update succeeds despite the caller not owning the record.

# Expected Behavior:  

The update is rejected because the caller does not own the target record and has no permission to modify it.

# Actual Behavior:  

The update succeeds. Confirmed locally as guest (user id 3) against admin-owned records: soft-deleting a non-public document and flipping an admin-owned document to public: true, each verified directly in the database. Both returned success: true to the guest caller. The same code path also has no ownership check for other state fields (closed, end, disable) and for arbitrary declared columns via the id fast path, but those were not separately reproduced.

# Screenshots:  

N/A

# System Information: 

Device (OS/Type): macOS
Browser (+ Version): Brave (Chromium)
CARE Software (Version/Branch): feat-83-recording

# Additional Information:  

The fix likely belongs in updateData, an ownership/permission check before updateById, applied consistently across all branches including the status-field fast path. Found while reproducing a recorderDelete review comment on #200.

theultimatewasteofspace:
Fixes #361, fixes #310. `updateData` in `AppSocket` updated rows by id with no ownership check, so any
logged-in user could modify another user's row through `appDataUpdate`.

Instead of adding a separate ownership check, this reuses the existing access logic:
`getFiltersAndAttributes` is split into `getReadFilter` and `getWriteFilter` over a shared
private implementation, and `assertWriteAccess` applies the write filter to the row being
written. The two paths differ in one flag: on the write path, being able to see a row never
implies permission to change it. `dev` is merged in (`0cd6af6d`), as the branch was 363
commits behind.

## New Dev Features

- `Socket.getReadFilter` and `Socket.getWriteFilter` over the shared private
  `#buildAccessFilter`, replacing `getFiltersAndAttributes`. On the write path,
  `publicTable`, `public` rows, shared rows from `getUserFilter` and column-only `accessMap`
  rights grant nothing.
- On the write path, `accessMap` table rules only grant write when `by` is `id`, meaning the
  user owns the referenced row in the other table (e.g. sessions in studies they own). Rules
  with any other `by` express membership (e.g. `study`'s rule via `study_session`) and grant
  read only. Without this, a participant with a session in a study could write to that study
  through `appDataUpdate`. Verified as guest on a study they only participate in: denied, while
  the study is still readable.
- `Socket.assertWriteAccess(tableName, id, options)` applies the write filter to a single
  row and throws `ACCESS_DENIED` if the row is not reachable. It rejects non-numeric ids,
  since Sequelize reads `{id: [1, 2]}` as `id IN (1, 2)`.
- `getUserFilter` now returns `{owned, shared}`, so the socket decides which applies and
  uses `shared` on reads only. Implemented in `template`, `assignment`, `ai_model`,
  `ai_hook`, `ai_hook_models` and `ai_budget`.
- Read and write access filters are documented in
  `docs/source/for_developers/backend/data_transfer.rst`.

## Improvements

- If no condition grants a user any row, access is denied instead of returning an
  unrestricted filter, and a warning is logged.
- `ai_model` and `ai_hook` returned only shared ids and relied on the generic `userId` rule,
  which `dev` no longer applies when a model defines `getUserFilter`. They now return
  `owned: {userId}` explicitly.
- `configuration.js` uses `assertWriteAccess` (and passes `options`) instead of checking
  `accessAllowed` alone, which only reported whether the user could reach any row.
- `assignmentAdd` goes through `addReviewerRequest`, which verifies the caller may manage the
  target study before delegating to `addReviewer`.
- `handleLimitations` was dead code and is removed.

## Bug Fixes

- Any logged-in user could update another user's row through `appDataUpdate`, including
  soft-deleting it (#361).
- `assignment_share` had no ownership check at all. `userId` on that table is the recipient,
  so the generic rule treated the recipient as the owner: a non-admin could repoint their own
  share row to any assignment, or create a share for themselves from scratch, both giving
  read access to arbitrary assignments. Fixed with
  `static foreignOwner = {column: "assignmentId", table: "assignment"}`, the mechanism
  `ai_model_share` and `ai_hook_share` already use.

## Known Limitations

- The fail-closed check also applies on reads, which is wider than suggested in review. An
  empty condition list means nothing grants the user a row on either path, and without it a
  future table with no `userId`, no `getUserFilter` and no public column would get an
  unrestricted read filter. It is close to unreachable on reads today and logs when it fires.
- A model returning `{}` from `getUserFilter` used to mean full row access and now grants
  nothing. No model does this and admins never reach that branch, so there is currently no
  way for a model to express full row access.
- Managing `assignment_share` rows now requires owning the assignment, so an admin who does
  not own it can no longer edit its shares. This matches the AI share tables.
- `assertWriteAccess` covers `appDataUpdate` and the configuration content update only.

## Future Steps

- Audit the remaining sockets that update rows by id and are not covered by
  `assertWriteAccess`.
- Decide whether models need an explicit way to declare full row access, now that `{}` no
  longer does it.

### DIFF
--- a/backend/webserver/Socket.js
+++ b/backend/webserver/Socket.js
@@ -597,6 +597,35 @@ module.exports = class Socket {
     }
 
     /**
+     * Throws unless the user may write to a specific row.
+     * Applies the write filter to the target row: if the row is not reachable
+     * under that filter, the user does not own it and the write is denied.
+     * @param {string} tableName The table being written to
+     * @param {number} id The id of the row being written
+     * @param {Object} options Handler options; options.transaction is passed through
+     *        so rows created earlier in the same transaction remain visible
+     * @returns {Promise<void>} Resolves when access is allowed
+     * @throws {TranslatableError} ACCESS_DENIED when the row is not writable by this user
+     */
+    async assertWriteAccess(tableName, id, options = {}) {
+        const {filter, accessAllowed} = await this.getWriteFilter(
+            this.userId, {id: id, deleted: false}, {}, tableName, this.rolesUpdatedAt
+        );
+        if (!accessAllowed) {
+            throw new TranslatableError("errors.permission.cannotUpdateOtherUserTable", {dataTable: tableName}, "ACCESS_DENIED");
+        }
+        const rows = await this.models[tableName].getAll({
+            where: filter,
+            attributes: ["id"],
+            transaction: options.transaction,
+        });
+        if (rows.length === 0) {
+            this.logger.warn("User with id " + this.userId + " tried to write row " + id + " in table " + tableName + " without access");
+            throw new TranslatableError("errors.permission.cannotUpdateOtherUserTable", {dataTable: tableName}, "ACCESS_DENIED");
+        }
+    }
+
+    /**
      * Row filters and column attributes for writing to a table.
      * Unlike the read path, publicTable and public rows grant nothing.
      * Callers must apply the returned filter to the row they intend to write —
--- a/backend/webserver/Socket.js
+++ b/backend/webserver/Socket.js
@@ -477,51 +477,6 @@ module.exports = class Socket {
     }
 
     /**
-     * Creates database filters according to limitations in the accessMap.
-     * @param {string} tableName The name of the table to create limitations for
-     * @param {Object} allFilter Starting filters
-     * @param {Object} accessMap AccessMap with limitations
-     * @param {Array<Object>} accessRights Access rights for the user
-     * @param {number} userId Id of user to check limitations for
-     * @returns {Object} array of limitation filters
-     */
-    handleLimitations(tableName, allFilter, accessRights, accessMap, userId) {
-
-
-        let filteredAccessMap = accessMap
-            .flatMap(a => {
-                const idField = a.access.target || 'id'; // Use 'target' if available, fallback to 'id'
-                return a.limitation
-                    ? {[idField]: {[Op.in]: [...new Set(a.limitation)]}}
-                    : null;
-            })
-            .filter(Boolean);
-
-
-        if (this.models[tableName].autoTable && 'userId' in this.models[tableName].getAttributes()) {
-            // Ensure we always include the 'userId' condition
-            filteredAccessMap = filteredAccessMap.concat([{userId: userId}]);
-        }
-
-        const limitedFilter = {
-            [Op.and]: [
-                allFilter,
-                {
-                    [Op.or]: filteredAccessMap
-                }
-            ]
-        };
-
-        const columns = [...new Set(
-            accessRights
-                .filter(a => a.columns)
-                .flatMap(a => a.columns)
-        )];
-
-        return {filter: limitedFilter, columns};
-    }
-
-    /**
      * Modifies allFilter and allAttributes according to user rights in the table.
      * @param {number} userId User ID to check the rights for
      * @param {Object} allFilter Starting filters
--- a/backend/webserver/Socket.js
+++ b/backend/webserver/Socket.js
@@ -477,15 +477,20 @@ module.exports = class Socket {
     }
 
     /**
-     * Modifies allFilter and allAttributes according to user rights in the table.
+     * Builds row filters and column attributes according to user rights in the table.
+     * Shared implementation for the read and write paths — call getReadFilter or
+     * getWriteFilter instead of using this directly.
      * @param {number} userId User ID to check the rights for
      * @param {Object} allFilter Starting filters
      * @param {Object} allAttributes Starting attributes
      * @param {string} tableName The table to check the rights for
      * @param {Date} rolesUpdatedAt Date of the last role update of the user
-     * @returns {Object} modified filters and attributes + whether access is allowed
+     * @param {boolean} publicGrantsAccess Whether publicTable and public rows grant access.
+     *        True when reading (public data is readable by anyone); false when writing,
+     *        since being visible must never imply permission to modify.
+     * @returns {Promise<Object>} modified filters and attributes + whether access is allowed
      */
-    async getFiltersAndAttributes(userId, allFilter, allAttributes, tableName, rolesUpdatedAt) {
+    async #buildAccessFilter(userId, allFilter, allAttributes, tableName, rolesUpdatedAt, publicGrantsAccess) {
         const accessMap = this.server.db.models[tableName]['accessMap'] || [];
         const filteredAccessMap = await this.filterAccessMap(accessMap, userId, rolesUpdatedAt);
         const relevantAccessMap = filteredAccessMap.filter(item => item.hasAccess);
@@ -493,7 +498,7 @@ module.exports = class Socket {
         const model = this.models[tableName];
         const hasModelUserFilter = typeof model.getUserFilter === "function";
         const isAdmin = await this.isAdmin(userId, rolesUpdatedAt);
-        const isPublicOrAdmin = isAdmin || model.publicTable;
+        const isPublicOrAdmin = isAdmin || (publicGrantsAccess && model.publicTable);
         const hasAccessRules = accessMap.length > 0;
         const hasUserIdAttribute = model.autoTable && 'userId' in model.getAttributes();
 
@@ -516,7 +521,8 @@ module.exports = class Socket {
             }
 
             // --- Public rows: always visible regardless of ownership or access rights ---
-            if ('public' in model.getAttributes()) {
+            // Read-only: a public row may be seen by anyone, but not written by anyone.
+            if (publicGrantsAccess && 'public' in model.getAttributes()) {
                 rowVisibilityConditions.push({public: true});
             }
 
@@ -575,6 +581,36 @@ module.exports = class Socket {
             };
         }
         return {filter: allFilter, attributes: allAttributes, accessAllowed: true};
+    }
+
+    /**
+     * Row filters and column attributes for reading a table.
+     * @param {number} userId User ID to check the rights for
+     * @param {Object} allFilter Starting filters
+     * @param {Object} allAttributes Starting attributes
+     * @param {string} tableName The table to check the rights for
+     * @param {Date} rolesUpdatedAt Date of the last role update of the user
+     * @returns {Promise<Object>} modified filters and attributes + whether access is allowed
+     */
+    async getReadFilter(userId, allFilter, allAttributes, tableName, rolesUpdatedAt) {
+        return await this.#buildAccessFilter(userId, allFilter, allAttributes, tableName, rolesUpdatedAt, true);
+    }
+
+    /**
+     * Row filters and column attributes for writing to a table.
+     * Unlike the read path, publicTable and public rows grant nothing.
+     * Callers must apply the returned filter to the row they intend to write —
+     * checking accessAllowed alone is not enough, as it only reports whether the
+     * user can reach any row in the table at all.
+     * @param {number} userId User ID to check the rights for
+     * @param {Object} allFilter Starting filters
+     * @param {Object} allAttributes Starting attributes
+     * @param {string} tableName The table to check the rights for
+     * @param {Date} rolesUpdatedAt Date of the last role update of the user
+     * @returns {Promise<Object>} modified filters and attributes + whether access is allowed
+     */
+    async getWriteFilter(userId, allFilter, allAttributes, tableName, rolesUpdatedAt) {
+        return await this.#buildAccessFilter(userId, allFilter, allAttributes, tableName, rolesUpdatedAt, false);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


backend/webserver/Socket.js
```diff
@@ -561,7 +521,8 @@ module.exports = class Socket {
             }
 
             // --- Public rows: always visible regardless of ownership or access rights ---
-            if ('public' in model.getAttributes()) {
+            // Read-only: a public row may be seen by anyone, but not written by anyone.
+            if (publicGrantsAccess && 'public' in model.getAttributes()) {
```
dennis-zyska, line 525: Dropping the public row condition here does not cover `getUserFilter`, called just below. `template.getUserFilter` returns `{public: true, type in [4, 5]}` for non-admins, and that condition goes into the same OR list, so any user can still edit or delete another user's public template through `appDataUpdate`. Pass the write mode into `getUserFilter` as a new argument so it can leave the public branch out when writing.

backend/webserver/Socket.js
```diff
@@ -622,6 +583,65 @@ module.exports = class Socket {
...
+     * under that filter, the user does not own it and the write is denied.
+     * @param {string} tableName The table being written to
+     * @param {number} id The id of the row being written
+     * @param {Object} options Handler options; options.transaction is passed through
+     *        so rows created earlier in the same transaction remain visible
+     * @returns {Promise<void>} Resolves when access is allowed
+     * @throws {TranslatableError} ACCESS_DENIED when the row is not writable by this user
+     */
+    async assertWriteAccess(tableName, id, options = {}) {
+        const {filter, accessAllowed} = await this.getWriteFilter(
```
dennis-zyska, line 611: A column-only `accessMap` right still grants full row access on the write path: it carries no `limitation`, so `hasUnlimitedRights` becomes true and `fullRowAccess` is set. Every user holds `frontend.dashboard.studies.view.userPublicInfo` on the `user` table, so `{id: someUserId, deleted: true}` passes this check and soft-deletes that account. Either gate `hasUnlimitedRights` on `publicGrantsAccess` as well, or add it to Known Limitations, since the description says ownership is now checked on every path.

backend/webserver/Socket.js
```diff
@@ -622,6 +583,65 @@ module.exports = class Socket {
...
+        );
+        if (!accessAllowed) {
+            throw new TranslatableError("errors.permission.cannotUpdateOtherUserTable", {dataTable: tableName}, "ACCESS_DENIED");
+        }
+        const rows = await this.models[tableName].getAll({
+            where: filter,
+            attributes: ["id"],
+            transaction: options.transaction,
+        });
+        if (rows.length === 0) {
```
dennis-zyska, line 622: This only checks that the filter matched at least one row, not that it matched the row being written. Sequelize reads `{id: [1, 2]}` as `id IN (1, 2)`, so `appDataUpdate` with `data: {id: [ownRowId, otherUserRowId], deleted: true}` matches the user's own row, passes here, and `updateById` then runs `WHERE id IN (1, 2)` and writes both. Reject a non-scalar `id` at the top of `assertWriteAccess`.

backend/webserver/sockets/configuration.js
```diff
@@ -39,7 +39,7 @@ class ConfigurationSocket extends Socket {
     }
 
     // Check if user has access to update this configuration
-    const { accessAllowed } = await this.getFiltersAndAttributes(
+    const { accessAllowed } = await this.getReadFilter(
```
dennis-zyska, line 42: Known Limitations covers the permission bug here, but the fix is now one line in this PR: replace this block with `await this.assertWriteAccess("configuration", configurationId, options);`. `configuration` is a `publicTable`, so `getReadFilter` returns `accessAllowed: true` for every logged-in user and this check never fails.

dennis-zyska (changes requested):

backend/webserver/sockets/configuration.js
theultimatewasteofspace: Done. Replaced with assertWriteAccess. Verified as guest (denied) and admin (still works).

backend/webserver/Socket.js
theultimatewasteofspace: Fixed, hasUnlimitedRights is now gated on publicGrantsAccess. Verified: {id: 4, deleted: true} as guest is denied, and the admin delete flow in Users dashboard still works.

There might be a consequence of this. On the write path a non-admin now fails at the early-denial branch for user, since user has no userId column and there's no ownership condition left once unlimited rights are off. I think no UI path hits that currently as there's no profile page and Users.vue is admin only. but it's a behaviour change beyond the attack itself, so please let me know if we should rather it kept an ownership fallback for tables keyed on id.

backend/webserver/Socket.js
theultimatewasteofspace: Fixed, assertWriteAccess now rejects a non-scalar id before building the filter. Verified: {id: [1, 2], deleted: true} as guest is denied, scalar writes to an owned row still pass.

backend/webserver/Socket.js
theultimatewasteofspace: youre right and confirmed, template.getUserFilter returns {public: true, type in [4,5]} for non admins, which goes into the same OR list, so templates are still writable by anyone.

But before I add the parameter: Socket.js:531 calls model.getUserFilter(userId) with one argument, but template.getUserFilter(userId, isAdmin) expects two, so isAdmin is undefined there today and the non admin branch runs for everyone, admins included. assignment.getUserFilter(userId) takes one and is async. Adding write mode as a third parameter means fixing that positional call too, which changes template read behaviour as a side effect. Do you want that fixed first as its own commit, or should getUserFilter take an options object instead?

theultimatewasteofspace:
Two more findings while working on this, both fixed here rather than in separate PRs:

configurationUpdate checks the accessAllowed boolean from the access function, which can only be false when a user can't reach any row in the table at all. configuration is a publicTable, so it's always true and the check never fires. Reproduced as guest and overwrote an admin-owned configuration. Now uses assertWriteAccess.

assignmentAdd takes studyId straight from the payload with no check anywhere in the file. Reproduced as guest, created a study session on an admin's study, and with enableEmailNotification set it would also send emails on the owner's behalf. The socket now goes through addReviewerRequest, which verifies the caller may manage the study before delegating; createAssignment still calls addReviewer directly since it just created that study itself.

Both were the same mistake as #361 in different handlers, I think the ownership check may eventually belong in MetaModel?

backend/webserver/Socket.js
dennis-zyska: The guest account has no rights on the `user` table, so it stops at the early denial branch (`accessRights.length === 0`) rather than at the new gate. A `user` role account holds `frontend.dashboard.studies.view.userPublicInfo`, so that branch does not run, and `rowVisibilityConditions` then stays empty: no `userId` column, no public condition on the write path, no `getUserFilter`, and both rights are column only. An empty list skips `if (!fullRowAccess && rowVisibilityConditions.length > 0)`, so no filter is applied and `{id: <otherUserId>, deleted: true}` still goes through.

An ownership fallback would not help here, since the empty list is itself the problem. Deny instead, right after the `if (!fullRowAccess) { ... }` block:

```js
if (!publicGrantsAccess && !fullRowAccess && rowVisibilityConditions.length === 0) {
    return {filter: allFilter, attributes: allAttributes, accessAllowed: false};
}
```

That covers every table with no `userId` column and no `getUserFilter`, not just `user`.

backend/webserver/Socket.js
dennis-zyska: Line 531 is inside `if (!fullRowAccess)`, and `fullRowAccess` starts as `isPublicOrAdmin`, so admins never reach that call. It only runs for non-admins, where a falsy `isAdmin` is already the correct value, so the missing argument has no effect today. The signature does not need to change.

Instead of passing a mode down into the model, `getUserFilter` can return its two conditions separately and let `Socket.js` decide which of them apply, the same way it already treats its own ownership and public conditions:

```js
// template.js
static getUserFilter(userId, isAdmin) {
    const {Op} = require("sequelize");
    const ofType = (cond) => isAdmin ? cond : {[Op.and]: [cond, {type: {[Op.in]: [4, 5]}}]};
    return {owned: ofType({userId}), shared: ofType({public: true})};
}

// Socket.js:531
const {owned, shared} = await model.getUserFilter(userId);
if (owned) {
    rowVisibilityConditions.push(owned);
}
if (shared && publicGrantsAccess) {
    rowVisibilityConditions.push(shared);
}
```

This needs the fail-closed check from the other thread to land first, otherwise a model that returns only `shared` leaves the write path with an empty condition list again. The existing `{}` case drops out: `template` and `assignment` are the only implementations and neither returns empty, so an empty return is better handled by that denial than by granting full row access.

The model then only describes its own rows, and the write rule stays in one place. `assignment.getUserFilter` becomes `owned: {userId}` and `shared: {id: {[Op.in]: assignedIds}}`, which also settles the sharing question: a shared assignment is readable but not writable, unless you deliberately move that condition into `owned`.

theultimatewasteofspace: Done, implemented as you described. getUserFilter returns {owned, shared} and
Socket.js pushes owned always, shared only when publicGrantsAccess. Signature
unchanged, dev already passes isAdmin at that call site. The fail-closed check
landed first, so the {} case drops out as you said.

One thing to point out. I merged dev first (it was 363 commits
behind) and dev added four more implementations, so the split covers six models:
template, assignment, ai_model, ai_hook, ai_hook_models, ai_budget. ai_model and
ai_hook only returned shared ids and relied on the generic userId rule, which dev
no longer applies when a model defines getUserFilter, so they now return
owned: {userId} explicitly.

I tested manually, as a non-admin, denied on another user's document, on a public
template, on an assignment shared with me, and on an AI model shared with me,
allowed on my own rows, admin unaffected, reads unchanged (public templates still
visible in the store and in the Public Templates modal).

backend/webserver/Socket.js
theultimatewasteofspace: Implemented in c67b29b4, right after the if (!fullRowAccess) { ... } block:

if (!fullRowAccess && rowVisibilityConditions.length === 0) {
    this.logger.warn(...);
    return {filter: allFilter, attributes: allAttributes, accessAllowed: false};
}

One difference from your snippet: I left out !publicGrantsAccess, so it denies on
the read path as well. An empty condition list means nothing grants this user a row
either way, and it is close to unreachable on reads since a column-only right already
sets fullRowAccess there. If you prefer write-only, let me know and I'll change it

Verified as guest: {id: 4, deleted: true} on user is denied and the row is unchanged,
the admin Users dashboard still lists and deletes. No "without any row access"
warnings during normal page loads.

theultimatewasteofspace:
Found this while testing, separate from the review comments, and fixed it in 1c4e2daf.

assignment_share had no ownership check: userId on that table is the recipient, so the
generic ownership rule treated the recipient as the owner. As a non-admin I could, via
appDataUpdate, repoint my own share row to any assignment and create a share for myself
from scratch, both giving read access to arbitrary assignments through
assignment.getUserFilter.

Fix is one line, reusing what ai_model_share and ai_hook_share already do:
static foreignOwner = {column: "assignmentId", table: "assignment"}, so MetaModel.add
and updateById check the parent assignment's owner.

Side effect worth knowing: managing shares now requires owning the assignment, so an
admin who does not own it can no longer edit its shares. That matches how the AI share
tables already behave. Verified both attacks are denied and that the normal share flow
in the Assignments dashboard still works.

PR #375, opened 2026-09-15
Fix 374 assignment start end time

mohammadsherif0:
## Summary
Stop create and edit from saving an assignment or study whose start is on or after its end. The form shows a toast and the model rejects the row.

## Bug Fixes
- Assignment create/edit (`AssignmentModal`) toasts `errors.assignment.startAfterEnd` and does not emit `appDataUpdate` when both times are set and start is not before end.
- Assignment model `assertStartBeforeEnd` on create, and on update only if `start` or `end` changed, so close/disable/delete still work.
- Same rule on studies: Coordinator toast when `table === "study"`, and `Study.assertStartBeforeEnd` on create and on start/end updates. That also covers save-as-template, import, and bulk/single study assignment, which copy the template through `study.add`.
- English and German strings for `errors.assignment.startAfterEnd` and `errors.studies.startAfterEnd`.


backend/db/models/assignment.js
```diff
@@ -135,6 +136,29 @@ module.exports = (sequelize, DataTypes) => {
...
 			return filter;
 		}
+		/**
+		 * Refuse inverted or zero-length assignment windows. Either date may be null.
+		 *
+		 * @param {object} assignment - Sequelize assignment instance or plain row with start/end.
+		 * @returns {void}
+		 * @throws {TranslatableError} If both dates are set and start is not before end.
+		 */
+		static assertStartBeforeEnd(assignment) {
```
dennis-zyska: `assertStartBeforeEnd` here is the same code as `Study.assertStartBeforeEnd` in `study.js`. Only the error key is different. Please move it into one helper in `backend/utils/` that takes the error key, and call it from both models.

frontend/src/basic/dashboard/Coordinator.vue
```diff
@@ -223,6 +223,15 @@ export default {
         return;
       }
 
+      if (this.table === "study" && this.isStartOnOrAfterEnd(this.data)) {
```
dennis-zyska: `Coordinator` is a generic component, so it should not have a rule for one table like `study`. The new `Study` model hook already rejects this save, and the failure branch below shows the same toast with that error. Please remove this check and the `isStartOnOrAfterEnd` method.

dennis-zyska (changes requested):

dennis-zyska (approved):

PR #376, opened 2026-09-18
Fix 366 publish assessment grades for hook studies

### LINKED ISSUE #366: [BUG] Publish assessment grades and hashes are not published correctly
# Describe the Bug:  

Publish Assessment does not publish grades or hashes correctly. Sending usernames and passwords still works.

# Reproducibilty:  

- [x] Yes 
- [ ] No 
- [ ] Sometimes 

Not specified.

1. Open Publish Assessment from the study view.
2. Select an assessment configuration, workflow step, and closed study sessions.
3. On the confirmation step, review the hashes (by studies or by sessions).
4. Choose Moodle as the publishing method, pick the course/assignment, and publish (including review URL / hashes if that option is used).

# Expected Behavior:  

Assessment scores should be converted to the Moodle assignment grade scale and written to the selected assignment. Session hashes / review URLs should be published correctly (as Moodle feedback when that option is enabled, and in the Hash column on CSV export). Username and password publishing should keep working, as it already does.

# Actual Behavior:  

Usernames and passwords are sent as expected. Grades are not applied as intended. Hashes are not published correctly.

# Screenshots:  

None provided.

# System Information: 

- **Device (OS/Type):** Omarchy 4.0.2 (Linux 7.1.9-arch1-2, x86_64, Hyprland/Wayland) 
- **Browser (+ Version):** Chrome 152.0.7977.82 
- **CARE Software (Version/Branch):** 1.1.0 / dev (12708bda) 

# Additional Information:  
credential publish path (usernames and passwords), which still succeeds. The failure is in Publish Assessment: grade values and hashes.

akash9676:
## Summary
Publish Assessment and project grade export were reading only `assessment_result` and flattening scores as `{ Criterion: { currentScore } }`. Hook studies store `aiHook_{id}` (or `{name}_{skill}_assessment`) in other shapes, so CSV, Moodle grades, and the grades ZIP came out as zeros. This PR looks up those keys, flattens the stored value through `scoresFromStoredValue` in `assessment-score`, and prefills Moodle connection fields from settings.

## Improvements
- Moodle course ID, API key, and API URL are copied from `rpc.moodleAPI.*` when a parent form starts with `{}` (`MoodleOptions.vue`). Assignment ID is still picked per publish.
- `assessment-score` exports `scoresFromStoredValue` for NLP lists, numeric maps, and saved criterion state. Publish Assessment and `export.js` both import it. Tests cover the list and map cases.

## Bug Fixes
- Publish Assessment CSV and Moodle grades now use hook/NLP keys as well as `assessment_result`, including trigger rows with a null session (`PublishAssessmentModal.vue`).
- The Moodle step binds `moodleOptions: {}` so the form actually renders.
- Project grade export queries `aiHook_%` and `*_assessment` as well as `assessment_result`, groups duplicate keys, and prefers a non-zero saved result over the hook row (`export.js`).

## Known Limitations
- N/A

## Future Steps
- N/A


backend/webserver/routes/export.js
```diff
@@ -340,19 +340,27 @@ module.exports = function (server) {
...
+     * @returns {Object} Flat map of criterion name to score
      */
-    function parseAssessmentState(rawAssessmentState) {
-        try {
-            const parsed = JSON.parse(rawAssessmentState);
-            return parsed && typeof parsed === "object" ? parsed : {};
-        } catch (error) {
-            console.warn("Failed to parse assessment state:", error.message);
-            return {};
+    function pickScoresFromGradeRows(rows) {
```
dennis-zyska: `pickScoresFromGradeRows` is the same code as `pickScoresFromDocumentData` in `PublishAssessmentModal.vue`. Please move it into the shared `assessment-score` package (used by both frontend and backend) next to `scoresFromStoredValue`, with a JSDoc and a unit test, and call it from both places. That way the CSV and the ZIP export can't drift apart.

backend/webserver/routes/export.js
```diff
@@ -340,19 +340,27 @@ module.exports = function (server) {
...
+        for (const row of rows) {
+            const scores = scoresFromStoredValue(row.value);
+            if (!Object.keys(scores).length) continue;
+            if (row.key === ASSESSMENT_RESULT_KEY) {
+                savedScores = scores;
+            } else if (!Object.keys(hookScores).length) {
+                hookScores = scores;
+            }
+        }
+        if (Object.values(savedScores).some((value) => Number(value) !== 0)) {
```
dennis-zyska: A non-zero check doesn't tell you whether a human graded this. If a grader saves 0 points on every criterion, the export uses the AI score instead. The saved value already stores `isSaved` per criterion, so please prefer the saved row when any criterion has `isSaved === true`. The same check is in `PublishAssessmentModal.vue:882`.

frontend/src/components/dashboard/submission/PublishAssessmentModal.vue
```diff
@@ -776,19 +773,13 @@ export default {
...
-      // Find any configured NLP skill or AI hook.
       const svc = cfg.services.find((s) => s.skill || s.hookId) || cfg.services[0];
-
       return svc || null;
     },
-    /**
-     * Get assessment data key for a study step.
-     * Returns canonical AI/NLP keys, otherwise "assessment_result".
-     */
     getAssessmentDataKeys(studyStep) {
```
dennis-zyska: Please keep the JSDoc on `getAssessmentDataKeys` and update it instead of deleting it. It should say that the method now always adds `assessment_result` to the hook/NLP keys.

utils/modules/assessment-score/tests/assessment-score.test.js
```diff
@@ -1,4 +1,20 @@
-const { calculateAssessmentScore, buildScoresFromState } = require("../index");
+const { calculateAssessmentScore, buildScoresFromState, scoresFromStoredValue } = require("../index");
+
+describe("scoresFromStoredValue", () => {
```
dennis-zyska: The new tests only cover arrays and plain maps. Please also test the string inputs (fenced JSON, double-encoded JSON) and the `{ output: ... }` wrapper, since those are the hook shapes this PR is fixing.

dennis-zyska (changes requested):

PR #384, opened 2026-09-23
Feat 275 placeholder options

mohammadsherif0:
## Summary
Prompt (type 8) placeholders can carry per-token `wordRange` and `pageRange` options so authors limit what text is sent to the model. Limits are applied when the prompt is resolved, from PDF.js page lists at hook extract time, not as a second limiter in the hook InputMap.

## New User Features
- Type 8 placeholders `pdfText` and `submissionFiles` can take Word range and Page range; `editorText` is seeded with Word range only.
- The Placeholders sidebar collects From/To before insert, for example `~pdfText[1]{pageRange:1}~` or `~editorText[1]{wordRange:1-500}~`.
- Word range and page range can be set on the same token. Pages are applied first, then words (`{pageRange:1,wordRange:1-200}` is the first 200 words of page 1).
- Omit options to send the full extract. Unknown or invalid option values are listed in the existing invalid-placeholder warning.

## New Dev Features
- Shared `utils/modules/placeholder-tokens` parses option braces, validates ranges, and applies `applyTextRangeLimit` (pages then words). Covered by unit tests including reversed option-key order, overflow clamp, and `{ pages, pageCount }` replacement (not `[object Object]`).
- `placeholder.placeholderOptions` holds option definitions; migration seeds `wordRange` / `pageRange` on the type 8 text placeholders.
- Hook PDF extract returns `{ pages, pageCount }` from PDF.js `numPages`. `resolveServiceInput` keeps that object when the mapping is `pdf` so the resolver can slice pages.
- Saving a template rejects a repeated option name on one token (e.g. two `wordRange` values).

## Improvements
- Silent 2000-character caps on `editorText` and `submissionFiles` are removed; omit options to send the full extract.
- Frontend and backend consume the shared placeholder-tokens module instead of duplicated token helpers.
- Developer docs cover option syntax, apply order, TeX/zip as one implicit page, and where type 8 resolve runs (`hook.js` vs `templateResolve`).

## Bug Fixes
- Template editor preview state is declared in `data` so preview mode does not fail on first use.

## Known Limitations
- A single number `N` always means the first N units, not unit N only. `pageRange:2` is pages 1–2; page 2 only is `pageRange:2-2`.
- Page slicing needs a `pages` array from PDF.js at hook extract. Backend-only `templateResolve` without that object treats the extract as one page. TeX/zip `submissionFiles` text is also one implicit page.
- Extend placeholderOptions and resolver handling to other placeholders and additional option types beyond limits.

#275


Makefile
```diff
@@ -71,6 +71,7 @@ test-rpc: backend/node_modules/.uptodate $(UTILS_MODULES_UPTODATE)
 test-modules: $(UTILS_MODULES_UPTODATE)
 	cd utils/modules/editor-delta-conversion && npm run test:module -- tests/editor-delta-conversion.test.js
 	cd utils/modules/assessment-score && npm run test:module -- tests/assessment-score.test.js
+	cd utils/modules/placeholder-tokens && npm run test:module -- tests/placeholder-tokens.test.js
```
dennis-zyska: `placeholder-tokens` is missing from `UTILS_MODULES_UPTODATE` (line 58). So `make test-modules` never runs `npm install` for it, and this line fails on a fresh checkout because `jest` is not installed there. Please add `utils/modules/placeholder-tokens/node_modules/.uptodate` to that list.

backend/db/migrations/20260720121142-extend-placeholder-options.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+
+const wordRangeOption = {
+  name: 'wordRange',
+  label: 'Word range',
```
dennis-zyska: `label: 'Word range'` is stored as plain English, but the other placeholder fields use i18n keys (for example `templates.placeholders.labels.prompt.pdfText`). Please use keys like `templates.placeholders.options.wordRange` and show them with `translateMaybeKey` in `TemplateConfigurator.vue`.

backend/webserver/sockets/template.js
```diff
@@ -612,6 +613,13 @@ class TemplateSocket extends Socket {
         ids: duplicates.join(", "),
       });
     }
+    const optionDuplicates = getDuplicatePlaceholderOptionTokens(content);
+    if (optionDuplicates.length > 0) {
+      throw new Error(
```
dennis-zyska: The check above uses `TranslatableError`, but this one throws a plain English `Error`. Please add an `errors.templates.duplicatePlaceholderOptions` key to the en and de `errors.json` and throw `new TranslatableError("errors.templates.duplicatePlaceholderOptions", { tokens: optionDuplicates.join(", ") })`.

frontend/src/components/editor/sidebar/TemplateConfigurator.vue
```diff
@@ -18,6 +18,14 @@
         </ul>
         {{ $t("templates.placeholders.duplicateBracketIdsHelp") }}
       </div>
+
+      <div v-if="optionApplyWarnings.length > 0" class="alert alert-warning mb-3">
+        <strong>Notice:</strong> The placeholder was added, but these options were not included:
```
dennis-zyska: This notice and the new option controls use hardcoded English (`Notice:`, `Select option...`, `From`, `To`, `Add option`, and the warning texts in `collectSelectedOptions` and `validateTokenOptions`). The rest of this component uses `$t(...)`, for example `templates.placeholders.duplicateBracketIdsHelp` above. Please move these strings into `templates.json` (en and de).

frontend/src/components/editor/template/TemplateEditor.vue
```diff
@@ -139,8 +139,8 @@
         newLanguageModalMessage: "",
         beforeUnloadHandler: null,
         previewMode: false,
-        placeholderPreviewList: [],
         lastEditorHtml: "",
+        placeholderPreviewList: [],
```
dennis-zyska: This only moves `placeholderPreviewList: []`, which is already in `data` on `dev`. So the "preview state declared in data" bug fix from the PR description is not in this diff. Please revert this line, or add the real fix if something is still missing.

dennis-zyska (approved):

Issue #383, opened 2026-09-23
[BUG] imported prompt template is created without its text


mohammadsherif0:
# Describe the Bug:

Importing a prompt template creates the template in the list, but the prompt text is missing. The same Import and Export buttons are used for every template type, so email and document templates lose their text the same way.

The text is not on the template row. It is stored per language in template_content. Export built the file from the template list in the browser, which does not include those rows. Import then inserted only the template row.

# Reproducibilty:

- [x] Yes
- [ ] No
- [ ] Sometimes

# Steps to Reproduce the Behavior:

1. Open Dashboard, then Templates.
2. Add a template with type Prompt, open it, write some text, and leave the editor so the text is saved.
3. Export that template as JSON or YAML.
4. Import the file.
5. Open the imported template.

# Expected Behavior:

The imported template contains the same prompt text, including every saved language.

# Actual Behavior:

A new template row is created. The editor is empty.

PR #385, opened 2026-09-23
Fix 383 import templates

mohammadsherif0:
## Summary
Importing a content template created the row and left the saved text behind. Export and import now carry each saved language body, and a non-admin can choose a prompt template when creating one.

## New Dev Features
- `templateExport` and `templateImport` on the template socket. Export reads owned templates through `findOwnedWithContent`.
- The content-template user story and the templates component doc describe that round trip.

## Improvements
- `ExportFormatModal` can download rows passed in by the caller, so template export can save bodies that are not in the client store.

## Bug Fixes
- Dashboard template import writes the language bodies from the file onto the new template. The new row belongs to the importer and is not published.
- Export of one template, or of every template on the user's list, includes those bodies. Another user's public template is not included.
- The frontend template type list includes type 8, so the add form offers Prompt to a non-admin.

## Known Limitations
- Import reads JSON and YAML. CSV and TXT stay on the export dialog, as they do for tags.
- An open draft in `template_edit` is not in the file. `sourceId` is omitted. A `public` flag in the file is ignored.

#383


backend/webserver/sockets/template.js
```diff
@@ -902,6 +902,128 @@ class TemplateSocket extends Socket {
...
+    const defaultLanguage = data.defaultLanguage || rows[0]?.language || "en";
+    if (rows.length > 0 && !seen.has(defaultLanguage)) {
+      throw new TranslatableError("errors.templates.importDefaultLanguageMissing", { language: defaultLanguage });
+    }
+
+    for (const row of rows) {
+      await this.assertNoDuplicatePlaceholders(row.content, type, options);
+    }
+
+    const template = await Template.add({
```
dennis-zyska: `importTemplate` repeats most of `createTemplate`: the required-field check, the admin check for email types, and adding the template plus its content rows. Please move that into one model method, for example `Template.createWithContents(payload, contents, options)`, and call it from both handlers. Then a future rule change only has to be made once.

dennis-zyska (changes requested):

Issue #386, opened 2026-09-23
[BUG] student can delete a submission PDF from Documents


mohammadsherif0:
# Describe the Bug:

A student who uploads a PDF for a submission can later delete that same PDF from Documents. That file is the submission, so deleting it from Documents should not be possible.

# Reproducibilty:

- [x] Yes
- [ ] No
- [ ] Sometimes

# Steps to Reproduce the Behavior:

1. Sign in as a student.
2. Upload a PDF for an open assignment from Submissions.
3. Open Documents.
4. Find that PDF and delete it.

# Expected Behavior:

The submission PDF stays in Documents. The student has no way to delete it from there.

# Actual Behavior:

Documents offers delete on that PDF. Confirming the delete removes the document.

PR #388, opened 2026-09-23
Fix 386 submission doc delete

mohammadsherif0:
## Summary
A student can no longer delete a submission PDF from Documents. The file stays linked to the submission, and deleting the submission or the project still removes it.

## Bug Fixes
- Documents hides delete when the file has a `submissionId`.
- `Document.beforeUpdate` rejects a soft-delete of that file and a change to an existing `submissionId`. The error is `errors.documents.submissionFileNotDeletable` in English and German.
- The submission soft-delete cascade and the project cascade pass `allowSubmissionDocumentDelete`, so those deletes still remove the file. The submission cascade also passes `options.transaction`.

#386


backend/db/models/document.js
```diff
@@ -424,6 +424,20 @@ module.exports = (sequelize, DataTypes) => {
             modelName: 'document',
             tableName: 'document',
             hooks: {
+                beforeUpdate: (document, options) => {
+                    if (options.allowSubmissionDocumentDelete) {
```
dennis-zyska: Please pass the flag through `options.context` instead of a custom top-level Sequelize option. Sequelize may drop or overwrite unknown option keys. Read it here with `options.context?.allowSubmissionDocumentDelete`. In `project.js` and `submission.js`, pass `{ transaction, context: { allowSubmissionDocumentDelete: true } }`.

docs/source/for_developers/basics/user_stories.rst
```diff
@@ -486,6 +486,8 @@ Rename and Delete a Document
 
      - When I try to rename a document to an empty name, I see a validation error.
      - When I cancel the delete confirmation dialog, the document is not deleted.
+     - When the document is a submission file, Documents does not offer delete. A delete
+      does not remove the document.
```
dennis-zyska: This continuation line has one space too few. It has to line up with the text after `- ` (7 spaces). Otherwise Sphinx ends the list item here and warns about an unexpected unindent. Please indent it with 7 spaces.

dennis-zyska (changes requested):

